PR #33, opened 2016-06-07
Documents design

unbrice:
This proposes and documents new schemas for biggraphite.

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/33)

### DIFF
--- a/CASSANDRA_DESIGN.md
+++ b/CASSANDRA_DESIGN.md
@@ -1,0 +1,75 @@
+# Cassandra backend
+This document describes how we modeled a TSDB using Cassandra.
+
+------
+
+## data tables
+For the full SQL please see [cassandra.py](biggraphite/drivers/cassandra.py).
+
+### Primary key
+All rows in data table describe have time and metric UUID as their primary key, and values as columns.<br />
+We group related timestamps (`_ROW_SIZE_MS`) in the same row described by `time_start_ms` and using `time_offset_ms` to describe  a delta from it. This saves space in two ways:
+ - No need to repeat metric IDs on each row.
+ - The relative time offset is 4 bytes only when a timestamp would be 8.
+
+### Expiry of data
+TTLs take space and are only applied on compaction. Tweaking compaction is tricky and that makes it hard to recover space in emergency by expiring data early.<br />
+An alternative would be to have one table per day of expiry, but for 10 years of retentions that would be about 3650 tables.<br />
+So what we do is that we round up retention policy to the next power of two and create a table per such retention policy.
+
+At any time we have 11 tables, corresponding to 1 TTL: 4 days, 8 days, 16 days, ..., 2048 days.
+We write and read a timestamp in the table that best fit. If we assume there is about the same amount of data week-over-week, we store **50% more than instant deletion**.
+
+**Example:**
+On the 28th of December 2015, the following 10 tables exist: "2016-01-01", "2016-01-09", "2016-01-25", "2016-02-24", ..., "2019-08-11".
+
+| TTL       |       |  E    | X     | P     | I     | R     | Y     |       |
+|--:	    |:-:	|:-:	|:-:	|:-:	|:-:	|:-:	|:-:	|:-:	|
+| **8d** 	|Jan 01 |Jan 09 |       |       |       |       |       |       |
+| **16d**	|Jan 09 |Jan 09 |Jan 25 |Jan 25 |       |       |       | *...* |
+| **32d**   |Jan 25 |Jan 25 |Jan 25 |Jan 25 |Feb 24 |Feb 24 |Feb 24 | *...* |
+| *...*     | *...* | *...* | *...* | *...* | *...* | *...* | *...* | *...* |
+| **2048d**	|2019   |2019   |2019   |2019   |2019   |2019   |2019   | *...* |
+
+So a point written on 2015-12-28 with a desired TTL of 30 days should expire on 2016-01-27 but instead will expire on 2016-02-24 .
+
+------
+
+## metadata tables
+Metadata are stored in two different tables, one for metrics and one for directories.
+
+Advantages of having separate two tables (current solution):
+  - We expect them to end up with increasingly different metadata (quotas on directories and
+    whisper-header-like on metrics) and the semantic of a merged entry will become complex
+  - We suspect that implementing quota or cleaning of empty directories will be easier if
+    we can iterate on directories without filtering all metrics
+
+Advantages of using one table:
+  - We get batch transactions to update metadata at the same time as parent entries
+  - We could still build a materialized view of directories only to get performance
+    benefits above
+
+None of the points above is a strong decision factor, but merging things later is generally easier than untangling them so we kept them separate for now.
+
+
+### "metrics" metadata table
+**Metrics metadata are: a name, a map of config keys to json values and a UUID.**
+
+To resolve globs, we store metrics in a table that among other things contain 64 columns, one for each path component.
+
+The metric `a.b.c` is therefore indexed as:
+```python
+   name="a.b.c", config=...,
+   component_0="a", component_1="b", component_2="c", component_3="__END__",
+   component_4=NULL, component_5=NULL, ..., component_63=NULL
+```
+A SASI Index is declared on each of the component columns.<br />
+We use SASI because it knows to process multi-column queries without filtering. The way it works is by indexing tokens in a B+Tree for each column of each sstable. After finding the right places in the B+Trees, SASI picks the column with the least results, and merge the result together. It uses multiple lookups if one column is 100 times bigger than the other, otherwise it uses a merge join.
+
+SASI support for LIKE queries is limited to finding substrings. So when resolving a query like `a.x*z.c` we query for `a.*.c` and then do client-side filtering. <br />
+[See here for more details on SASI](https://github.com/apache/cassandra/blob/trunk/doc/SASI.md)
+
+### "directories" metadata table
+Similar to the metrics table, we create for parents of all metrics an entry in a directory table. It is used to implement the 'metric' API in graphite.
+
+Directories are implicitely created before metrics by the Accessor.create_metric() function. Because of this, it is possible to end up with an empty directory (if the program crashes). A cleaner job will be needed if they accumulate for too long.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,104 @@
+# How to contribute to BigGraphite
+
+First thanks :heart: :heart: :heart:
+
+This document explains how to contribute code or bug reports, but feel free to ask your questions on [Gitter](https://gitter.im/criteo/biggraphite) or our [mailing list](https://groups.google.com/forum/#!forum/biggraphite).
+
+
+## Did you find a bug?
+
+- Ensure the bug was not already reported by searching on GitHub under [Issues](https://github.com/criteo/biggraphite/issues).
+
+- If you're unable to find an open issue addressing the problem, [open a new one](https://github.com/criteo/biggraphite/issues/new). Be sure to include a *title and clear description*, as much relevant information as possible.
+
+## Code contributions
+
+See [README](README.md) for the code structure.
+
+### Style guide
+
+By convention:
+- This project follows (PEP 8)[https://www.python.org/dev/peps/pep-0008/]
+  and (PEP 257)[https://www.python.org/dev/peps/pep-0257/]
+- We indent by four space
+- We only import modules (`from os import path as os_path` not `from os.path import join`).
+
+### Did you write a patch that fixes a bug?
+
+- Check information on style above, `pylama tests biggraphite` will tell you about many violations
+- Open a new GitHub pull request with the patch.
+- Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
+
+### Do you intend to add a new feature or change an existing one?
+
+- Check information on style above.
+- Suggest your change in the [mailing list](https://groups.google.com/forum/#!forum/biggraphite) and start writing code.
+- Do not open an issue on GitHub until you have collected positive feedback about the change. GitHub issues are primarily intended for bug reports and fixes.
+
+### Test Environment
+
+```bash
+# Setup the virtualenv
+$ export BG_VENV=bg
+$ virtualenv ${BG_VENV}
+$ source bg/bin/activate
+
+# Install Graphite dependencies
+$ export GRAPHITE_NO_PREFIX=true
+$ pip install -r requirements.txt
+$ pip install -r tests-requirements.txt
+
+# Install Cassandra
+$ export CASSANDRA_VERSION=3.5
+$ wget "http://www.us.apache.org/dist/cassandra/${CASSANDRA_VERSION}/apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz"
+$ tar -xzf "apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz"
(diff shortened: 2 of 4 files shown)


iksaif:
First round of comments. *not* on the design yet

<details><summary><i>Previously, coveralls wrote…</i></summary>
> 
> [Coverage Status](https://coveralls.io/builds/6495844)
> 
> Coverage increased (+0.3%) to 93.142% when pulling **6bb34c2ec88b9b00606418d5349f42af823bd14d on unbrice:wip** into **a71d8fcd34a282c1d76decf7c0b2092b3b2b38b2 on criteo:master**.
> 
</details>


---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[CONTRIBUTING.md, line 2 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJg99ZwgPblxCfXL6Bx:-KJg99ZxiB0K9gWXy71U:37836832) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L2)):*
> ```Markdown
> # How to contribute to Ruby on Rails
> ```

ruby on rails ?

---

*[CONTRIBUTING.md, line 21 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJgAqZ9la00efxchdYj:-KJgAqZ9la00efxchdYk:-1530565135) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L21)):*
> ```Markdown
> - This project follows (PEP 8)[https://www.python.org/dev/peps/pep-0008/]
>   and (PEP 257)[https://www.python.org/dev/peps/pep-0257/]
> - We intend by four space
> ```

indent

---

*[CONTRIBUTING.md, line 23 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJgArpjl35uzT4qqGV5:-KJgArpjl35uzT4qqGV6:-1839709414) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L23)):*
> ```Markdown
> - We intend by four space
> - We only import modules (`from os import path as os_path` not `from os.path import join`).
> ```

say that one should run pylama ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33#-:-KJgAjm7dO1TzYR3YCwH:-1266672145)*

unbrice:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[CONTRIBUTING.md, line 2 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJg99ZwgPblxCfXL6Bx:-KJkldzKaK5z5ilRNsED:1749860100) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L2)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> ruby on rails ?
</details>
:-P Done.

---

*[CONTRIBUTING.md, line 21 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJgAqZ9la00efxchdYj:-KJkmdZsx_0To08MFLLZ:-499146009) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L21)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> indent
</details>
Done.

---

*[CONTRIBUTING.md, line 23 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJgArpjl35uzT4qqGV5:-KJknq3h4P53Oxc1hRYH:-499146009) ([raw file](https://github.com/criteo/biggraphite/blob/c66be891dc8d8d6d8df1237d2230847fae96fc9c/CONTRIBUTING.md#L23)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> say that one should run pylama ?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33)*

iksaif:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl1gq2WhkgxReNDNA_:-KJl1gq2WhkgxReNDNAa:-847139483) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
> ```Markdown
> 
> ### Expiry of data
> We use counters, so we cannot use TTL. Also TTLs are only applied on compaction and finding proper compaction configuration not to waste place would be tricky.<br />
> ```

Why do we *have* to use counters ? Looking at https://docs.datastax.com/en/cql/3.1/cql/cql_using/use_counter_t.html counters seem complicated and rather new (and thus probably fragile). It also seem to be a sharded counter where broken replication could easily break stuff.

Compared to that, just write the same value to two cassandra nodes seems more robust (even if we waste some performances on the reading path).

Remember that we want to optimize the write path, the read path will be cached most of the time.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl2U1ibdnjxCvjicag:-KJl2U1ibdnjxCvjicah:1199887340) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
> ```Markdown
> 
> ### Expiry of data
> We use counters, so we cannot use TTL. Also TTLs are only applied on compaction and finding proper compaction configuration not to waste place would be tricky.<br />
> ```

Why would compaction waste space ?

---

*[CASSANDRA_DESIGN.md, line 21 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3XJg8ZhEcDCH2vE9:-KJl3XJg8ZhEcDCH2vEA:-1885055865) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L21)):*
> ```Markdown
> 
> At any time we have 11 tables prefix, corresponding to 1 TTL: 4 days, 8 days, 16 days, ..., 2048 days.
> We write and read a timestamp in the table that best fit. If we assume there is about the same amount of data week-over-week, we store **50% more than instant deletion**.
> ```

So 50% more is basically also the worst case with compactions + ttl ?

---

*[biggraphite/driver/interface.py, line 27 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3rF-1ezKxD1ZuTQO:-KJl3rF-1ezKxD1ZuTQP:-1987612466) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/biggraphite/driver/interface.py#L27)):*
> ```Python
> import abc
> 
> class Driver(object):
> ```

I'm generally fine with all this, it doesn't really belong here though 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33)*

unbrice:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl1gq2WhkgxReNDNA_:-KJug0f5KYbprAOlxKnN:-549748577) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> Why do we *have* to use counters ? Looking at https://docs.datastax.com/en/cql/3.1/cql/cql_using/use_counter_t.html counters seem complicated and rather new (and thus probably fragile). It also seem to be a sharded counter where broken replication could easily break stuff.
> 
> Compared to that, just write the same value to two cassandra nodes seems more robust (even if we waste some performances on the reading path).
> 
> Remember that we want to optimize the write path, the read path will be cached most of the time.
</details>
Great question. Counters were already there in 0.8. [Since 2.1 it's no longer a sharded counter](http://www.datastax.com/dev/blog/whats-new-in-cassandra-2-1-a-better-implementation-of-counters). Now the node takes a local lock on the row and reads/add/write - with a cache. This is basically what we'd be implementing in Python, except I expect it to be faster in Java, we reuse code and avoid sharding by hand.

[These benchmarks](http://riptano.github.io/cassandra_performance/graph/graph.html?stats=6553.low_contention_no_cache_CL_quorum_3_28.json&metric=op_rate&operation=counterwrite&smoothing=1&xmin=0&xmax=140.03&ymin=0&ymax=51540.5) from [CASSANDRA-6553](https://issues.apache.org/jira/browse/CASSANDRA-6553) indicate that in case of low contention (10K counters in different partitions accessed by 50 threads) and with CL.quorum, authors were achieving 44k writes per second *without caching*.

So I expect performance to be many order of magnitudes above what we need. And it's very cheap to implement.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl2U1ibdnjxCvjicag:-KJwnTgNj73UPxVgyivz:-1127533170) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> Why would compaction waste space ?
</details>
[Because compaction with DTCS happens every so often](https://labs.spotify.com/2014/12/18/date-tiered-compaction/). After 4H for data written more than 1H ago, after 16H for >4H ago, after 3 days for >16H, 12D for >3D and so forth. Not only does old data remain for an extra time proportional to how long it has existed (as with proposal) but furthermore this behaves very poorly when TTL of different rows are different and interleaved in one SSTable as the "drop the sstable" idea described does not work.
Also we mix application of retention policies with SSTable compaction which makes all sort of tuning more complex.

---

*[CASSANDRA_DESIGN.md, line 21 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3XJg8ZhEcDCH2vE9:-KJwr4Qcgu_ThjAY14yg:-1462734851) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L21)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> So 50% more is basically also the worst case with compactions + ttl ?
</details>
My understanding is that compaction + TTL has a worst case of 300% with default min_threshold=4.

---

*[biggraphite/driver/interface.py, line 27 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3rF-1ezKxD1ZuTQO:-KJwre7GnkE2KrlXnKnV:-241364814) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/biggraphite/driver/interface.py#L27)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> I'm generally fine with all this, it doesn't really belong here though 
</details>
ACK, will make it part of the driver refacto review.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33)*

iksaif:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl2U1ibdnjxCvjicag:-KK7lqD4Q1orjn18JlSS:970424995) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
<details><summary><i>Previously, unbrice (Brice Arnould) wrote…</i></summary>
> [Because compaction with DTCS happens every so often](https://labs.spotify.com/2014/12/18/date-tiered-compaction/). After 4H for data written more than 1H ago, after 16H for >4H ago, after 3 days for >16H, 12D for >3D and so forth. Not only does old data remain for an extra time proportional to how long it has existed (as with proposal) but furthermore this behaves very poorly when TTL of different rows are different and interleaved in one SSTable as the "drop the sstable" idea described does not work.
> Also we mix application of retention policies with SSTable compaction which makes all sort of tuning more complex.
</details>
That's with DateTieredCompactionStrategy, which isn't the default and that we may not want to use.
Other databases usually only have SizeTieredCompactionStrategy. Cassandra also have LeveledCompactionStrategy.

Also, as far as I know, these only affect minor compactions. Major compactions are still triggered periodically (a period which is probably between 1 and 7 days by default).

---

*[CASSANDRA_DESIGN.md, line 21 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3XJg8ZhEcDCH2vE9:-KK7mutXKn6pr_TJQEtZ:1780260796) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L21)):*
<details><summary><i>Previously, unbrice (Brice Arnould) wrote…</i></summary>
> My understanding is that compaction + TTL has a worst case of 300% with default min_threshold=4.
</details>
How do you get 300% ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33)*

unbrice:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[CASSANDRA_DESIGN.md, line 16 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl2U1ibdnjxCvjicag:-KLpmpfRrc89RkvVvIeA:314253233) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L16)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> That's with DateTieredCompactionStrategy, which isn't the default and that we may not want to use.
> Other databases usually only have SizeTieredCompactionStrategy. Cassandra also have LeveledCompactionStrategy.
> 
> Also, as far as I know, these only affect minor compactions. Major compactions are still triggered periodically (a period which is probably between 1 and 7 days by default).
</details>
Discussed IRL.

---

*[CASSANDRA_DESIGN.md, line 21 \[r2\]](https://reviewable.io:443/reviews/criteo/biggraphite/33#-KJl3XJg8ZhEcDCH2vE9:-KLpmst44IHMyjDsRsU6:160356787) ([raw file](https://github.com/criteo/biggraphite/blob/d20cefa8bfca362485ba3a7400bf42297b146377/CASSANDRA_DESIGN.md#L21)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> How do you get 300% ?
</details>
4-1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/33)*

unbrice:
Discussions have been had, let's turn this into a proper doc.

PR #52, opened 2016-07-22
Set TTLs as a stopgap measure

unbrice:
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/52)

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -172,7 +172,10 @@ class _CassandraAccessor(bg_accessor.Accessor):
         self.__insert_statement = self.__session.prepare(
             "INSERT INTO datapoints "
             "(metric, time_step_ms, time_start_ms, time_offset_ms, value, count)"
-            " VALUES (?, ?, ?, ?, ?, ?);")
+            " VALUES (?, ?, ?, ?, ?, ?)"
+            # One week TODO: this is a stopgap until we implement table-based TTLs
+            " USING TTL 604800;"
+        )
         self.__insert_statement.consistency_level = cassandra.ConsistencyLevel.ONE
         self.__select_statement = self.__session.prepare(
             "SELECT time_start_ms, time_offset_ms, value, count FROM datapoints"


iksaif:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[biggraphite/drivers/cassandra.py, line 177 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/52#-KNWLpL93ylUQznS-1zq:-KNWLpL93ylUQznS-1zr:372971370) ([raw file](https://github.com/criteo/biggraphite/blob/adb0b6a83308ee3cb6bf81041b2929b504dbb28a/biggraphite/drivers/cassandra.py#L177)):*
> ```Python
>             " VALUES (?, ?, ?, ?, ?, ?)"
>             # One week TODO: this is a stopgap until we implement table-based TTLs
>             " USING TTL 604800;"
> ```

can you set the proper ttl based on the retention policy instead ? (and possibility only set it if some kind of constent or conf is set to true ?)

(even if we're not all convinced that this would work, I'd like to keep the option to enabled it easily for experiments, it's fine if the option name contains 'BROKEN' or 'EXPERIMENTAL')

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/52)*

unbrice:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[biggraphite/drivers/cassandra.py, line 177 \[r1\]](https://reviewable.io:443/reviews/criteo/biggraphite/52#-KNWLpL93ylUQznS-1zq:-KNWlq4N1K09A1m7x46O:-1069612053) ([raw file](https://github.com/criteo/biggraphite/blob/adb0b6a83308ee3cb6bf81041b2929b504dbb28a/biggraphite/drivers/cassandra.py#L177)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary>
> can you set the proper ttl based on the retention policy instead ? (and possibility only set it if some kind of constent or conf is set to true ?)
> 
> (even if we're not all convinced that this would work, I'd like to keep the option to enabled it easily for experiments, it's fine if the option name contains 'BROKEN' or 'EXPERIMENTAL')
</details>
Will, do are you OK with having this in the meanwhile? It avoids me from removing it from the patch stack and doesn't hurt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/52)*

iksaif:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/52#-:-KNWo07F0MoxatxuCY1U:1417088721)*

PR #82, opened 2016-08-22
cassandra: change time_offset_ms to offset (which unit is stage.preci…

iksaif:
…sion_ms)

time_offset_ms was limited to 46 days (which doesn't work well when
you store >1000 points with a resolution of one day in a single row).

To overcome this limitation, change offset to be related to stage.precision_ms
which allows us to store up to 4B points in a single row (which we probably
won't do, and we will use the high bits for other purposes later).

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/82)

### DIFF
--- a/CASSANDRA_DESIGN.md
+++ b/CASSANDRA_DESIGN.md
@@ -14,9 +14,9 @@ For the full CQL please see [cassandra.py](biggraphite/drivers/cassandra.py).
 
 ### Primary key
 All rows in data table describe have time and metric UUID as their primary key, and values as columns.<br />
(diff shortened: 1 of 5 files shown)


wdauchy:
:lgtm_strong: 

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/82#-:-KPqMlENXyiHDG267vDw:b-uwkowd)*

PR #84, opened 2016-08-23
plugins: create cache and accessor lazilly

iksaif:
No need to create them before we use them. This will also prevent
creating the cache as root before we start using it as 'graphite'.

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/84)


wdauchy:
:lgtm_strong: 

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/84#-:-KPslHfcXaCWiEqpAz4Z:b-uwkowd)*

PR #104, opened 2016-09-01
downsampling: don't re-write stage0 points multiple times

iksaif:
Downsampler.feed() should not return already emitted stage0 points
else we end up write 20 points each time. Also rename 'raw' to 'stage0'.

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/104)

### DIFF
--- a/biggraphite/accessor.py
+++ b/biggraphite/accessor.py
@@ -267,6 +267,11 @@ class Stage(object):
     def __str__(self):
         return self.as_string
 
+    def __repr__(self):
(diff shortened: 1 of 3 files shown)


wdauchy:
:lgtm_strong: 

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/104#-:-KQ_db4BGZleY3Ch8gv3:b-uwkowd)*

PR #138, opened 2016-10-04
graphite: do not keep un-connected accessors

iksaif:
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/138)

### DIFF
--- a/biggraphite/plugins/graphite.py
+++ b/biggraphite/plugins/graphite.py
@@ -142,8 +142,12 @@ class Finder(object):
         """Return an accessor."""
         if not self._accessor:
             from django.conf import settings as django_settings
-            self._accessor = graphite_utils.accessor_from_settings(django_settings)
-            self._accessor.connect()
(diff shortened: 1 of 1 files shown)


biggraphite/plugins/graphite.py
```diff
@@ -142,8 +142,9 @@ def accessor(self):
         """Return an accessor."""
         if not self._accessor:
             from django.conf import settings as django_settings
-            self._accessor = graphite_utils.accessor_from_settings(django_settings)
-            self._accessor.connect()
+            accessor = graphite_utils.accessor_from_settings(django_settings)
+            accessor.connect()
```
rveznaver: Could you comment this line explaining that the connect on the accessor might error out and therefore we never return an unconnected one here?

iksaif: done

rveznaver:
+1

PR #139, opened 2016-10-05
accessor: do not create empty directories for double-dots

iksaif:
Double dots ('..') has the same meaning as '//' in a filesystem
so we should not create empty directories in this case.

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/139)


rveznaver:
+1

Issue #69, opened 2016-07-28
drivers: reduce write load


unbrice:
When we expire datapoints (most writes), we update consecutive points within one row. By using [unlogged batch](http://docs.datastax.com/en/cql/3.1/cql/cql_reference/batch_r.html), we can update all points in one Cassandra update.

We should not however put the different stage level in a batch. Instead we can skip updating retention levels if the value did not change.

iksaif:
http://docs.datastax.com/en/cql/3.1/cql/cql_using/useBatch.html

> Batches are often mistakenly used in an attempt to optimize performance. Unlogged batches require the coordinator to manage inserts, which can place a heavy load on the coordinator node. If other nodes own partition keys, the coordinator node needs to deal with a network hop, resulting in inefficient delivery. Use unlogged batches when making updates to the same partition key.


We still need to reduce the amount of downsampled points being written.

iksaif:
I ended doing that some time ago

PR #147, opened 2016-10-06
Bulk inserts for aggregates in Cassandra

hlerebours:
Might be a step towards reducing write load: https://github.com/criteo/biggraphite/issues/69

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/147)

### DIFF
--- a/biggraphite/accessor.py
+++ b/biggraphite/accessor.py
@@ -27,6 +27,8 @@ import math
 import re
 import threading
 
+from biggraphite.drivers import _downsampling
+
 
 class Error(Exception):
     """Base class for all exceptions from this module."""
@@ -562,10 +564,13 @@ class Accessor(object):
 
     __metaclass__ = abc.ABCMeta
 
+    max_requests_per_connection = None
+
     def __init__(self, backend_name):
         """Set internal variables."""
         self.backend_name = backend_name
         self.is_connected = False
+        self._downsampler = _downsampling.Downsampler(accessor=self)
 
     def __enter__(self):
         """Call connect()."""
--- a/biggraphite/drivers/_downsampling.py
+++ b/biggraphite/drivers/_downsampling.py
@@ -19,8 +19,76 @@ from __future__ import print_function
 
 import array
 import math
+import Queue
+import threading
+
 
 _NaN = float("NaN")
+
+_MAX_WRITE_DELAY = 5 * 60  # in seconds
+
+
+class _Writer(threading.Thread):
+    def __init__(self, accessor):
+        super(_Writer, self).__init__()
+        self._accessor = accessor
+        self._queue = Queue.Queue()
+        self._must_stop = threading.Event()
+        self._avg_sleep = 0
+
+    def soon(self, metric, downsampled, on_done=None):
+        self._accessor.insert_downsampled_points_async(metric, downsampled, on_done)
+
+    def later(self, metric, downsampled, on_done=None):
+        if not self.__started.is_set():
+            self.start()
+        self._queue.put_nowait((metric, downsampled, on_done))
+
+    def stop(self):
+        self._must_stop.set()
+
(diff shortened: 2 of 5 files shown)


biggraphite/accessor.py
```diff
@@ -562,10 +564,13 @@ class Accessor(object):
 
     __metaclass__ = abc.ABCMeta
 
+    max_requests_per_connection = None
```
iksaif: This is cassandra specific and shouldn't be moved here.

biggraphite/accessor.py
```diff
@@ -562,10 +564,13 @@ class Accessor(object):
...
+
     def __init__(self, backend_name):
         """Set internal variables."""
         self.backend_name = backend_name
         self.is_connected = False
+        self._downsampler = _downsampling.Downsampler(accessor=self)
```
iksaif: Same thing, not all drivers will use the software downsampler, some might later have database primitives to do that in the database directly

biggraphite/drivers/_downsampling.py
```diff
@@ -19,9 +19,77 @@
...
+            request_count = 0
+            for metric, on_done, aggregated in per_metric.itervalues():
+                downsampled = [
+                    (timestamp, value, count, stage)
+                    for (timestamp, stage), (value, count) in aggregated.iteritems()
+                ]
+                self.soon(metric, downsampled, on_done)
+                request_count += len(downsampled)
+
+            if max_requests and request_count > max_requests:
```
iksaif: You don't need to do that, you just need to make sure, when you start a loop from scratch, that you'll have enough time to write everything

biggraphite/drivers/_downsampling.py
```diff
@@ -19,9 +19,77 @@
 
 import array
 import math
+import Queue
+import threading
+
 
 _NaN = float("NaN")
 
+_MAX_WRITE_DELAY = 5 * 60  # in seconds
```
iksaif: Could this be a parameter of the writer thread ?

biggraphite/drivers/_downsampling.py
```diff
@@ -19,9 +19,77 @@
...
+        There's a dedicated Event to stop this -otherwise infinite- loop.
+        """
+        max_requests = self._accessor.max_requests_per_connection
+
+        while True:
+            if self._must_stop.wait(timeout=self._avg_sleep):
+                break
+
+            per_metric = {}
+            while not self._queue.empty():
```
iksaif: You'll end up using a ton of memory by doing that. I suggest you store points in one or multiple dicts directly (to allow easy override)

biggraphite/drivers/_downsampling.py
```diff
@@ -19,9 +19,77 @@
...
+import Queue
+import threading
+
 
 _NaN = float("NaN")
 
+_MAX_WRITE_DELAY = 5 * 60  # in seconds
+
+
+class _Writer(threading.Thread):
```
iksaif: The writer should probably not be in _downsampling.py (the role of _downsampling.py is to downsample points only). You can put that in _background_writer.py and have a self._background_writer in the accessors (or possibly have a Mixin that adds this capability)

biggraphite/drivers/memory.py
```diff
@@ -45,7 +44,6 @@ def __init__(self):
             sortedcontainers.SortedDict)
         self._name_to_metric = {}
         self._directory_names = sortedcontainers.SortedSet()
-        self.__downsampler = _downsampling.Downsampler()
```
iksaif: do not remove it, this is explicitely added to memory and cassandra because at some point some drivers won't have the use for a "soft" downsampler and will be able to do that in the database directly.

iksaif:
superseeded by #162

PR #150, opened 2016-10-07
bgutil list

rudy-6-4:
you can ignore accessor and command_du, requirements, it's part of previous merge request

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/150)

### DIFF
--- a/biggraphite/cli/command_list.py
+++ b/biggraphite/cli/command_list.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+# Copyright 2016 Criteo
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""List Command."""
+
+from __future__ import print_function
+
+from biggraphite.cli import command
+
+
+def list_metrics(accessor, pattern):
+    """Return the list of metrics corresponding to pattern. Exit with error message if None.
+
+    Args:
+        - accessor, Accessor, a connected accessor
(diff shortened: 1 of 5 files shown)


rudy-6-4:
https://github.com/criteo/biggraphite/issues/90

biggraphite/cli/command_list.py
```diff
@@ -0,0 +1,63 @@
...
+from biggraphite.cli import command
+
+
+def list_metrics(accessor, pattern):
+    """Return the list of metrics corresponding to pattern. Exit with error message if None.
+
+    Args:
+        - pattern, string, e.g. my.metric.a or my.metric.**.a
+    """
+    accessor.connect()
```
iksaif: it's kind of weird to depend on list_metric to call accessor.connect(). Could you keep that at the beginning of the command, and change the docstring to specify what you need a connected accessor ?

rudy-6-4: yes, that should be done in bgutils.main if you accept that all commands do the connect (even if they don't need)
otherwise, each command must do its own connect

iksaif: Yep, each command should start by doing a connect() if they are going to use the accessor. Some commands don't need to call connect() because they don't use it

rudy-6-4: ok

tests/test_bgutil.py
```diff
@@ -15,15 +15,30 @@
...
+from biggraphite.accessor import Metric, Retention
 
 
 class TestBgutil(bg_test_utils.TestCaseWithFakeAccessor):
 
-    def test_run(self):
-        bgutil.main(['--driver=memory', 'read', 'foo'])
+    metrics = ['metric1', 'metric2']
+
+    @patch('biggraphite.drivers.memory._MemoryAccessor.get_metric')
```
iksaif: Why do we need mocking ? Since you're in a TestCaswWithFakeAccessor with the memory driver you can easily insert metrics for testing directly.

rudy-6-4: Because I didn't want to have side effet, but ok it's better to use the memory driver. Can I reset easily the content of the memory for other tests?

iksaif:
Could you rebase and make sure that the check runs ?

rudy-6-4:
Yes

PR #153, opened 2016-10-07
Automatic stage selection from metric timestamp for bgutil read

rveznaver:
This PR should also save one fallback object creation request for the graphite plugin.

This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/153)

### DIFF
--- a/biggraphite/accessor.py
+++ b/biggraphite/accessor.py
@@ -447,6 +447,23 @@ class Retention(object):
         # There is always at least one stage.
         return self.stages[-1]
 
+    def align_time_window(self, start_time, end_time, now):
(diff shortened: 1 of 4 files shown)


biggraphite/accessor.py
```diff
@@ -447,6 +447,23 @@ def find_stage_for_ts(self, searched, now):
         # There is always at least one stage.
         return self.stages[-1]
 
+    def align_ts(self, start_time, end_time, now):
```
iksaif: align_time_window ?

iksaif: Also could you add a test for this new method ?

rveznaver: Done

Issue #154, opened 2016-10-07
graphite_utils.py: __filter_metric() doesn't work with more than one {}


iksaif:
We should at least be able to support something like a.{b,c}.{d,e}.{f,g}. The current code is limited to a single list of variants.

dpanth3r:
Pull request in progress :-)

iksaif:
Arg... I made a fix in the train but I failed to push it :/ hopefully we
have something similar

dpanth3r:
Well then, review time!

PR #155, opened 2016-10-07
Proper glob post-filtering

dpanth3r:
- Fixes memory driver behaviour
- Cleaned tests up

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/155)

### DIFF
--- a/biggraphite/drivers/memory.py
+++ b/biggraphite/drivers/memory.py
@@ -17,15 +17,13 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import collections
-import fnmatch
-import re
 import uuid
 
 import sortedcontainers
 
 from biggraphite import accessor as bg_accessor
+from biggraphite import glob_utils as bg_glob
 from biggraphite.drivers import _downsampling
-
 
 OPTIONS = {}
 
@@ -118,13 +116,9 @@ class _MemoryAccessor(bg_accessor.Accessor):
 
     @staticmethod
     def __glob_names(names, glob):
-        dots_count = glob.count(".")
-        glob_re = re.compile(fnmatch.translate(glob))
-
-        return [
-            name for name in names
-            if name.count(".") == dots_count and glob_re.match(name)
-        ]
+        results = bg_glob.glob(names, glob)
+        results.sort()
+        return results
 
     def glob_metric_names(self, glob):
         """See the real Accessor for a description."""
@@ -156,8 +150,8 @@ class _MemoryAccessor(bg_accessor.Accessor):
             # A row is time_base_ms, time_offset_ms, value, count
             row = self.Row(ts * 1000.0, 0, float(points[ts]), 1)
             rows.append(row)
-        query_results = [(True, rows)]
 
+        query_results = [(True, rows)]
         time_start_ms = int(time_start) * 1000
         time_end_ms = int(time_end) * 1000
         return bg_accessor.PointGrouper(
--- a/biggraphite/glob_utils.py
+++ b/biggraphite/glob_utils.py
@@ -1,0 +1,219 @@
+#!/usr/bin/env python
+# Copyright 2016 Criteo
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


dpanth3r:
Among other things, closes #154

dpanth3r:
For potential reviewer questions: I moved some features from graphite_utils to utils because it kind of made sense to me, AND it allowed to use the glob helpers from the memory driver (graphite_utils used to include utils, which would create an import cycle).

iksaif:
So, that was my version of the same thing: https://github.com/iksaif/biggraphite/commit/8028bd9677ff30478e7f534681d2c888e0efaf46

graphite-utils and utils where separated on purpose because we didn't want too many graphite specific things to leak into the common part (to make it easier to have opentsdb/influxdb compatibility layer later).

The idea was that the accessor would support only wildcards, that the anything smarter would be done somewhere else.

dpanth3r:
I would tend to say your suggested implementation is less understandable, and I believe mine will probably be a bit more efficient because it only uses a single regexp instead of as many as there are variants (which in the end makes little sense, apart maybe from the fact that it hides that between the fnmatch API).

I don't really understand your last point, because I did not move the smart stuff anywhere else. Unless you mean that about the in-memory driver, in which case I did it because it was already doing semi-smart things.

dpanth3r:
There is now a new more correct version of the glob_to_regex feature, improving on fnmatch. I have also moved the glob-related code outside of graphite_utils, and updated the pre-filtering code of memory in the process to be glob/fnmatch agnostic (well, ok, it still tries to detect whether there is a globbing expression in components, but it has to!). This latest version of the patch includes additional tests.

biggraphite/drivers/memory.py
```diff
@@ -118,14 +116,26 @@ def create_metric(self, metric):
...
-        glob_re = re.compile(fnmatch.translate(glob))
-
-        return [
-            name for name in names
-            if name.count(".") == dots_count and glob_re.match(name)
+        dots = glob.count(".")
+        literal_glob_components = [
+            (index, component)
+            for (index, component) in enumerate(glob.split("."))
+            if not bg_glob._is_graphite_glob(component)
```
iksaif: Can we avoid *graphite* specific thing in the drivers ?
In this case:
* basic wildcard globbing should be done in the driver
* advanced integration specific globbing (graphite) should be callbed by the driver

Suggestion: add a specific method for wildcard globbing in bg_glob that could be callbed by drivers who can't do wildcards themselves

biggraphite/utils.py
```diff
@@ -48,11 +48,11 @@ class ConfigError(Error):
 def accessor_from_settings(settings):
     """Get Accessor from configuration.
 
-    Args:
```
iksaif: The convention we use is really "Args"

dpanth3r: heh.

biggraphite/drivers/memory.py
dpanth3r: I'll add a helper in glob utils then, I knew this would come up :-)

dpanth3r: Done.

biggraphite/glob_utils.py
```diff
@@ -0,0 +1,204 @@
...
+      A tuple:
+        First element: sorted list of Cassandra metrics matched by the glob.
+        Second element: sorted list of Cassandra directories matched by the glob.
+    """
+    if not _is_valid_glob(graphite_glob):
+        # TODO(d.forest): should we instead raise an exception?
+        return ([], [])
+
+    glob_re = re.compile(_glob_to_regex(graphite_glob))
+    accessor_components = _graphite_glob_to_accessor_components(graphite_glob)
```
iksaif: I think there still a misunderstanding, in my mind we should have:
* graphite_glob(), used only by the graphite plugin: supports graphite thingies.
* glob(), used by the drivers when the database can't do it: supports star and globstar

Graphite specific things should *stay* in the graphite plugin scope and not leak further.

dpanth3r: Updated! Added globstar support to glob, too

PR #156, opened 2016-10-08
Fix globbing

iksaif:
See #154

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/156)


dpanth3r:
I have to disagree with what seems more complicated to me than just building one regex. Adding to this is the fact that fnmatch will actually build a regex every time you call filter, which makes this even more complex behind the scenes.

iksaif:
I'm merging this to have a quick/low-impact fix. I'm ok for it to be reverted by yours when it's ready.

PR #164, opened 2016-10-26
Add test for timestamp update whilst time travelling

rveznaver:
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/164)

### DIFF
--- a/tests/test_metadata_cache.py
+++ b/tests/test_metadata_cache.py
@@ -16,6 +16,7 @@
 from __future__ import print_function
 
 import unittest
+from freezegun import freeze_time
 
 from biggraphite import test_utils as bg_test_utils
 from biggraphite import metadata_cache as bg_metadata_cache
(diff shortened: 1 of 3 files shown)


tests/test_metadata_cache.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import print_function
 
 import unittest
+from freezegun import freeze_time
```
iksaif: I believe we already have mock, can we do the same thing with mock ?

rveznaver: Probably, but the examples I have seen so far are way more complicated than freezegun

tests/test_metadata_cache.py
```diff
@@ -128,6 +129,36 @@ class TestDiskCache(CacheBaseTest, bg_test_utils.TestCaseWithFakeAccessor):
 
     CACHE_CLASS = bg_metadata_cache.DiskCache
 
+    def _get_metric_timestamp(self):
+        """Get the metric timestamp directly from DiskCache."""
+        with self.metadata_cache._DiskCache__env.begin(
+            self.metadata_cache._DiskCache__metric_to_metadata_db, write=False
+        ) as txn:
+            payload = txn.get(_TEST_METRIC.name)
+        return payload.rsplit(self.metadata_cache._METRIC_SEPARATOR, 1)[1]
```
iksaif: Please add a function in the DiskCache that does the split and use it from the tests :)

PR #165, opened 2016-10-26
Add method to clean cache from expired metrics

rveznaver:
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/165)

### DIFF
--- a/biggraphite/metadata_cache.py
+++ b/biggraphite/metadata_cache.py
@@ -119,6 +119,11 @@ class Cache(object):
         return metric
 
     @abc.abstractmethod
+    def clean(self):
+        """Clean the cache from expired metrics."""
+        pass
+
+    @abc.abstractmethod
     def repair(self, start_key=None, end_key=None, shard=0, nshards=1):
         """Remove spurious entries from the cache.
 
@@ -193,6 +198,10 @@ class MemoryCache(Cache):
         """Put metric in the cache."""
         if metric:
             self.__cache[metric_name] = metric
+
+    def clean(self):
+        """Automatically cleaned by cachetools."""
+        pass
 
     def repair(self, start_key=None, end_key=None, shard=0, nshards=1):
         """Remove spurious entries from the cache."""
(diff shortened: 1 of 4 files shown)


biggraphite/metadata_cache.py
```diff
@@ -402,6 +411,23 @@ def __split_payload(self, payload):
...
+    def clean(self):
+        """Remove all expired metrics.
+
+        Note: This will also remove metrics without a timestamp or
+              whose timestamp value is not a digit.
+        """
+        with self.__env.begin(self.__metric_to_metadata_db, write=True) as txn:
+            cursor = txn.cursor()
+            for key, value in cursor:
+                timestamp_str = self.__split_payload(value)[-1]
```
iksaif: _, _, timestamp

biggraphite/metadata_cache.py
```diff
@@ -402,6 +411,23 @@ def __split_payload(self, payload):
...
+                if not timestamp_str.isdigit() or \
+                   int(timestamp_str) < (int(time.time()) - int(self.__ttl)):
```
iksaif: clean should take a time_fn=time.time and at the beginning of the loop you should add now = time_fn()

biggraphite/metadata_cache.py
```diff
@@ -432,7 +458,7 @@ def repair(self, start_key=None, end_key=None, shard=0, nshards=1):
                 metric = self._accessor.get_metric(key)
                 expected_value = self.__value_from_metric(metric) if metric else None
 
-                if self.__split_payload(value)[0:2] !=\
+                if self.__split_payload(value)[0:2] != \
```
iksaif: uuid, metadata, _ = split() ?

biggraphite/metadata_cache.py
rveznaver: That would fail if there is no timestamp (e.g. previous version of the database)

biggraphite/metadata_cache.py
rveznaver: good point, completely forgot about that

PR #167, opened 2016-10-28
metadata_cache: fix DiskCache.has_metric()

iksaif:
Before it would return True when a miss was being cached.

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/167)


iksaif:
@rveznaver

rveznaver:
Here's a llama, there's a llama, and another little llama
Fuzzy llama, funny llama, llama llama duck

So yeah... pylama is complaining :)

rveznaver:
:+1:

PR #173, opened 2016-11-09
Initial implementation of query optimization for Cassandra

dpanth3r:
Initial work for #169 

Summary:
* Adds a (simplistic) graphite pattern parser to our glob utils that can be reused
* Makes use of said parser in the Cassandra driver implementation to take advantage of prefix queries (because we are currently using prefix SASI indexes) on metric name components whenever possible

Warning:
* This has been tested quickly, but not thoroughly benchmarked yet

Future work:
* Add brace selector parsing, and use it to generate several queries using the generated values whenever possible (e.g. a{b,c,d}.b would generate 3 queries, for ab.b / ac.b / ad.b, instead of “a%.b” + potentially costly post-filtering)
* Caveat: beware of situations with character selectors inside the brace selector! (-> naive implementation will not be correct!)

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/173)

PR #175, opened 2016-11-10
cassandra: don't use LIKE queries

iksaif:
See #174

This is a quick fix before we have a fully working #173

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/175)

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -875,7 +875,9 @@ class _CassandraAccessor(bg_accessor.Accessor):
         Returns:
           Query string that is ready to be executed.
         """
-        if "*" in components:
+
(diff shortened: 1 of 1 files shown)


mcoolive:
I see no regression. BUT, the generated query contains (for example): " AND component_10 = '__END__' ALLOW FILTERING". Issue #174 says it is slow, so we should remove th test on '__END__'.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/175#-:-KWVsLtimOjTv6sfVSg_:b-c95vyp)*

iksaif:
True, this has been fixed by https://github.com/criteo/biggraphite/commit/7622b9cf86d4cedda36a04da6fbd8b7417ab010e
I had to bypass the review process because it was affecting the stability of the system in preprod and no reviewers were available :/

PR #188, opened 2016-11-23
Update TRAVIS in order to test Java tools

mcoolive:
- Specify the version of Python
- Specify the version of the JDK, and install it
- Add execution of existing Java tests. Run once to save time (on the
slave that checks pyjama).

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/188)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,46 @@
 language: python
+python:
+  - 2.7
+jdk:
+  - oraclejdk8
 
 install:
-  - pip install tox coveralls
+  ### Java: We download now some dependencies to limit the number of messages later
+  - if $JAVA; then mvn dependency:resolve-plugins dependency:resolve -B -V ; fi
+  ### Python build dependencies
+  - if [ -n "$TOXENV" ]; then pip install tox coveralls; fi
 script:
-  - tox
+  ### Java
+  - if $JAVA; then mvn verify -B -V ; fi
+  ### Python
+  - if [ -n "$TOXENV" ]; then tox; fi
 after_success:
-  - coveralls
+  ### Python
+  - if [ -n "$TOXENV" ]; then coveralls; fi
 
-# As of 2016-04 Travis provides Java 7 with precise so we use trusty
+### sudo is required by pip
 sudo: required
+
+### As of 2016-11, when Sudo is required, java8 is removed (WHY?)
 dist: trusty
+addons:
+  apt:
+    packages:
+      - oracle-java8-installer
+
 env:
+  global:
+    - JAVA=false # Default
+    ### As of 2016-04 Travis provides Cassandra 2.1 by default so we bundle our own
+    - CASSANDRA_VERSION=3.9
+    - CASSANDRA_HOME="${TRAVIS_BUILD_DIR}/apache-cassandra-${CASSANDRA_VERSION}/"
   matrix:
+    - JAVA=true
     - TOXENV=pylama
     - TOXENV=pypy-coverage
     - TOXENV=py27-coverage
-  global:
-    # As of 2016-04 Travis provides Cassandra 2.1 by default so we bundle our own
-    - CASSANDRA_VERSION=3.9
-    - CASSANDRA_HOME="${TRAVIS_BUILD_DIR}/apache-cassandra-${CASSANDRA_VERSION}/"
 
-# As of 2016-04 Travis provides Cassandra 2.1 by default so we bundle our own
+### As of 2016-04 Travis provides Cassandra 2.1 by default so we bundle our own
 before_install:
    - wget "http://www.us.apache.org/dist/cassandra/${CASSANDRA_VERSION}/apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz"
    - tar -xzf "apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,26 +1,44 @@
...
+  apt:
+    packages:
+      - oracle-java8-installer
+
 env:
   matrix:
-    - TOXENV=pylama
-    - TOXENV=pypy-coverage
-    - TOXENV=py27-coverage
+    - JAVA=true  TOXENV=pylama # We build Java once only
```
iksaif: Running the java tests in the pylama env is kind of weird.
What do you think about having a fourth case with JAVA=true and TOXENV being null ?
Also maybe we should make JAVA=false implicit.

mcoolive:
*[.travis.yml, line 33 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/188#-KXGb-wyJeVmZlLitrm0-r1-33:-KXfG4Haqq7qD_2pkTuO:b-pc6f13) ([raw file](https://github.com/criteo/biggraphite/blob/b4943724984b143351b5a12246d0356c0f227f96/.travis.yml#L33)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary><blockquote>

Running the java tests in the pylama env is kind of weird.
What do you think about having a fourth case with JAVA=true and TOXENV being null ?
Also maybe we should make JAVA=false implicit.
</blockquote></details>

Currently if JAVA is not specified in the travis matrix, JAVA is tested (so it acts as if true was the default value). I can change that easily.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/188)*

mcoolive:
*[.travis.yml, line 33 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/188#-KXGb-wyJeVmZlLitrm0-r1-33:-KXfJ7kGXgEt_jMUIo6Z:bqmxbtb) ([raw file](https://github.com/criteo/biggraphite/blob/b4943724984b143351b5a12246d0356c0f227f96/.travis.yml#L33)):*
<details><summary><i>Previously, mcoolive (Cyril Martin) wrote…</i></summary><blockquote>

Currently if JAVA is not specified in the travis matrix, JAVA is tested (so it acts as if true was the default value). I can change that easily.
</blockquote></details>

About the mangle Java/Pylama.
I agree mangle Java and Pylama doesn't make sense. First I look for a trick to execute JAVA tests once. I put them with pyjama (and not with other target) because it is the shortest build on travis, so it doesn't affect the build time too much (in fact, the build time increases by 30 sec because we install JDK8 on each env...) 

Separate them seams cleaner. But I don't know how to achieve that in a clean way. What do you mean by "TOXENV being null"? As far I understand, when TOXENV is undefined by travis, tox plays all envs declared in tox.ini. I could wrap the tox command in a "if $PYTHON" or "if ! $ JAVA", but it seams dirty and complex to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/188)*

iksaif:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[.travis.yml, line 33 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/188#-KXGb-wyJeVmZlLitrm0-r1-33:-KXfUEMNA22FP7o6-H0L:b-y1d5gm) ([raw file](https://github.com/criteo/biggraphite/blob/b4943724984b143351b5a12246d0356c0f227f96/.travis.yml#L33)):*
<details><summary><i>Previously, mcoolive (Cyril Martin) wrote…</i></summary><blockquote>

About the mangle Java/Pylama.
I agree mangle Java and Pylama doesn't make sense. First I look for a trick to execute JAVA tests once. I put them with pyjama (and not with other target) because it is the shortest build on travis, so it doesn't affect the build time too much (in fact, the build time increases by 30 sec because we install JDK8 on each env...) 

Separate them seams cleaner. But I don't know how to achieve that in a clean way. What do you mean by "TOXENV being null"? As far I understand, when TOXENV is undefined by travis, tox plays all envs declared in tox.ini. I could wrap the tox command in a "if $PYTHON" or "if ! $ JAVA", but it seams dirty and complex to me.
</blockquote></details>

What I mean is that this would look cleaner:
- TOXENV=pylama
- TOXENV=pypy-coverage
- TOXENV=py27-coverage
- JAVA=true

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/188)*

mcoolive:
*[.travis.yml, line 33 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/188#-KXGb-wyJeVmZlLitrm0-r1-33:-KXkXSA66nh993ukdpZT:b-896fix) ([raw file](https://github.com/criteo/biggraphite/blob/b4943724984b143351b5a12246d0356c0f227f96/.travis.yml#L33)):*
<details><summary><i>Previously, iksaif (Corentin Chary) wrote…</i></summary><blockquote>

What I mean is that this would look cleaner:
- TOXENV=pylama
- TOXENV=pypy-coverage
- TOXENV=py27-coverage
- JAVA=true


</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/188)*

iksaif:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/188#-:-KXkgz4WolEerTDfKgpX:bnfp4nl)*

PR #233, opened 2017-01-25
Cache and Creation

iksaif:
See commits

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/233)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,17 +6,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-- Prometheus support and additional metrics.
+- Prometheus support and additional metrics
+- Add support for multiple replicas/writers
+- Microbenchmarks
+- Asynchronous creation of metrics in carbon
+- Better bg-import-whisper / bg-generate-sstables + helper scripts
 
 ### Fixed
--
+- Multiple bugs with the disk cache
+- Multiple clean/repair bugs
 
 ### Changed
 - [Breaking] Cassandra schema as been changed to separate stage0 and aggregated
   metrics. There is no upgrade procedure as this is a pre-release.
-
-### Removed
--
+- [Breaking] `bgutil syncdb` needs to be run during initial install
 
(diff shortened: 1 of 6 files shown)


Thib17:
Reviewed 4 of 6 files at r1.
Review status: 4 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[biggraphite/metadata_cache.py, line 494 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/233#-KbQR1QX03RXLYe82gH6:-KbQR1QYUBv2bmxO-e3c:b-3h34rj) ([raw file](https://github.com/criteo/biggraphite/blob/28de9871728e28f589384f60804f73cec3f659dc/biggraphite/metadata_cache.py#L494)):*
> ```Python
>         for key, value in cursor:
>             count += 1
>             if count > 1:
> ```

why do we use a for if we only use the first element ? (count is never decrease)

---

*[biggraphite/plugins/carbon.py, line 227 at r1](https://reviewable.io:443/reviews/criteo/biggraphite/233#-KbQSBIfL4KjFrCG5dXP:-KbQSBIfL4KjFrCG5dXQ:b-3rquql) ([raw file](https://github.com/criteo/biggraphite/blob/28de9871728e28f589384f60804f73cec3f659dc/biggraphite/plugins/carbon.py#L227)):*
> ```Python
>         if existing_metric:
>             # The retention policy is different, update it.
>             log.creates("updating database metric %s (%s)" % (
> ```

Shouldn't we update the metric in addition to log it ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/233)*

iksaif:
Fixed both, added tests.

Thib17:
:lgtm:

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/criteo/biggraphite/233#-:-KbUCagGcYLrqcF14_IO:bnfp4nl)*

Issue #242, opened 2017-02-02
cassandra: make consistency configurable


iksaif:
For now let's just make it an option like the others. Each consistency should be configurable.

PR #248, opened 2017-03-06
cassandra: make consistency configurable #242

InformatiQ:
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/248)

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -78,6 +78,16 @@ DEFAULT_TIMEOUT_QUERY_UTIL = 120
 
 DIRECTORY_SEPARATOR = '.'
 
+consistency_name_to_value = cassandra.ConsistencyLevel.name_to_value
+
+
+def _consistency_validator(k):
+    if k in consistency_name_to_value.keys():
+        return k
+    else:
+        return None
+
+
 OPTIONS = {
     "keyspace": str,
     "contact_points": _utils.list_from_str,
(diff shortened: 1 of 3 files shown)


biggraphite/drivers/cassandra.py
```diff
@@ -95,6 +95,12 @@
     "enable_metrics": bool,
     "writer": lambda k: None if k is None else int(k),
     "replica": lambda k: 0 if k is None else int(k),
+    "meta_write_consistency": str,
```
iksaif: maybe add a proper validator here ?
The validator would take either a string, that it would convert to a ConsistencyLevel, or directly a consistency level (see https://github.com/datastax/python-driver/blob/5c5a53ef8de50a1da48964e9bc8e196f3780ec4b/cassandra/__init__.py#L95)

biggraphite/drivers/cassandra.py
```diff
@@ -208,14 +238,28 @@ class InvalidGlobError(InvalidArgumentError):
 
 # CONSISTENCY PARAMETERS
 # ======================
+
+CONSISTENCIES = {
```
iksaif: You can use https://github.com/datastax/python-driver/blob/5c5a53ef8de50a1da48964e9bc8e196f3780ec4b/cassandra/__init__.py#L95

biggraphite/drivers/cassandra.py
```diff
@@ -769,6 +816,12 @@ def __init__(self,
         self.__enable_metrics = enable_metrics
         self.__read_on_counter = 0
         self.__read_on_sampling_rate = 0.1
+        self._meta_write_consistency = cassandra.ConsistencyLevel.name_to_value[meta_write_consistency]
```
iksaif: self._meta_write_consistency = (
    cassandra.ConsistencyLevel.name_to_value[meta_write_consistency])

PR #268, opened 2017-04-24
Fix biggraphite install

Thib17:
IPython6 need python>3.3, so we force ipython<5

---
This change is [Reviewable](https://reviewable.io/reviews/criteo/biggraphite/268)


iksaif:
can you send a fix for Cassandra too ? :)

PR #298, opened 2017-10-10
tags: add boilerplate for tags

iksaif:
Adds a new empty tag db and make sure we are compatible with the upcomming
Graphite version which has some breaking changes.


iksaif:
cc: @DanCech

PR #306, opened 2017-11-17
cassandra: ignore timeout in background jobs

iksaif:


biggraphite/drivers/cassandra.py
```diff
@@ -2042,9 +2061,17 @@ def run(rows):
...
         token = start_token
         while token < stop_token:
-            rows = self._execute_metadata(select, (token,), DEFAULT_TIMEOUT_QUERY_UTIL)
-            if len(rows.current_rows) == 0:
-                break
+            try:
+                rows = self._execute_metadata(select, (token,), DEFAULT_TIMEOUT_QUERY_UTIL)
+
+                # Empty results means that we've reached the end.
+                if len(rows.current_rows) == 0:
```
wdauchy: if I am not wrong you can simply write
if not rows.current_rows:

biggraphite/drivers/cassandra.py
```diff
@@ -2001,11 +2010,22 @@ def clean(self, max_age=None, start_key=None, end_key=None, shard=1, nshards=0,
     def _prepare_background_request(self, query_str):
         select = self.__session_metadata.prepare(query_str)
         select.consistency_level = self._meta_background_consistency
-        select.retry_policy = cassandra.policies.DowngradingConsistencyRetryPolicy
+        select.retry_policy = cassandra.policies.DowngradingConsistencyRetryPolicy()
         select.request_timeout = DEFAULT_TIMEOUT_QUERY_UTIL
 
         return select
 
+    def _prepare_background_request_on_index(self, query_str):
```
Thib17: Should it be named ...on_index, or ...always_retry ?
I really don't see where is it specific to a query on index.

biggraphite/drivers/cassandra.py
```diff
@@ -2001,11 +2010,22 @@ def clean(self, max_age=None, start_key=None, end_key=None, shard=1, nshards=0,
...
+        select = self.__session_metadata.prepare(query_str)
+        select.retry_policy = bg_cassandra_policies.AlwaysRetryPolicy()
+        print(select)
```
Thib17: remove print()

biggraphite/drivers/cassandra.py
```diff
@@ -2001,11 +2010,22 @@ def clean(self, max_age=None, start_key=None, end_key=None, shard=1, nshards=0,
     def _prepare_background_request(self, query_str):
         select = self.__session_metadata.prepare(query_str)
         select.consistency_level = self._meta_background_consistency
-        select.retry_policy = cassandra.policies.DowngradingConsistencyRetryPolicy
+        select.retry_policy = cassandra.policies.DowngradingConsistencyRetryPolicy()
```
Thib17: What is the impact ?

biggraphite/drivers/cassandra.py
```diff
@@ -1807,9 +1808,17 @@ def map(self, callback, start_key=None, end_key=None, shard=1, nshards=0):
...
         token = start_token
         while token < stop_token:
-            rows = self._execute_metadata(select, (token,), DEFAULT_TIMEOUT_QUERY_UTIL)
-            if len(rows.current_rows) == 0:
-                break
+            try:
+                rows = self._execute_metadata(select, (token,), DEFAULT_TIMEOUT_QUERY_UTIL)
+
+                # Empty results means that we've reached the end.
+                if len(rows.current_rows) == 0:
```
wdauchy: same possible comment below

biggraphite/drivers/cassandra.py
iksaif: The previous thing was a bug

biggraphite/drivers/cassandra.py
iksaif: It prepares a background request that will use an index.

Index queries have specific needs.

biggraphite/drivers/cassandra.py
iksaif: done

biggraphite/drivers/cassandra.py
iksaif: I'd like to not change this code (I just moved it). I think I remember there was a reason for the len() but I'm not sure which one.

biggraphite/drivers/cassandra.py
```diff
@@ -1994,18 +2017,28 @@ def clean(self, max_age=None, start_key=None, end_key=None, shard=1, nshards=0,
         """
         super(_CassandraAccessor, self).clean(max_age, callback_on_progress)
 
-        self._clean_empty_dir(start_key, end_key, shard, nshards, callback_on_progress)
+        #self._clean_empty_dir(start_key, end_key, shard, nshards, callback_on_progress)
```
InformatiQ: why comment that line? if it is not needed delete it

biggraphite/drivers/cassandra.py
wdauchy: ok

biggraphite/drivers/cassandra.py
iksaif: typo

PR #310, opened 2017-12-14
Python 3 + Graphite 1.1.0

iksaif:


benchmarks/accessor_bench.py
```diff
@@ -48,21 +49,23 @@ def runTest(self):
         return True
 
 
+def _random_name(n):
+    return "".join(
+        [random.choice(string.digits+string.ascii_letters) for i in range(10)])
```
Thib17: /10/n/

iksaif:
Better now

PR #320, opened 2018-01-16
Fix listing of metrics & directories in CLI commands

rclaude:
…) to list metrics & directories.

graphite_glob() post filter accessor.glob_metric_names() to ensure output correctness
This fix affect the behavior of 'read', 'du', 'copy' and 'list' commands.
A better but more involved fix would be to fix glob_metric_names() cassandra implementation and trust its output but some code might rely on the current behavior.


iksaif:
I restarted the travis job :)

iksaif:
`biggraphite/cli/command_list.py:22:1: E302 expected 2 blank lines, found 1 [pep8]`

PR #321, opened 2018-01-23
Use glob_utils.graphite_glob() to list metrics & directories.

rclaude:
This fix affect the behavior of 'read', 'du', 'copy' and 'list' commands.
Currently accessor.glob_metric_names() is used but return in some cases invalid results
graphite_glob() post filter accessor.glob_metric_names() to ensure output correctness
A better but more involved fix would be to fix glob_metric_names() cassandra implementation and trust its output but some code might rely on the current behavior.

### DIFF
--- a/biggraphite/cli/command_list.py
+++ b/biggraphite/cli/command_list.py
@@ -17,18 +17,22 @@
 from __future__ import print_function
 
 from biggraphite.cli import command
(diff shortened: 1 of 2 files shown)


iksaif:
Added a commit on top of it, waiting for the tests to run.

PR #329, opened 2018-03-09
Add metrics

iksaif:
* Add metrics for a bunch of things, including operations that can often be slow.
* Make SYNC_EVERY_N_WRITES configurable to reduce write latency
* Fix cassandra connection leak when open fails


InformatiQ:
LGTM

PR #330, opened 2018-03-19
Fix non escaped char in the Lucene filter string

geobeau:


biggraphite/drivers/lucene.py
```diff
@@ -88,12 +86,13 @@ def translate_to_lucene_filter(components):
...
-        must_list.append(FIELD_MATCH_VALUE % ('parent', parent))
+        must_list.append(json.dumps(
+            {"field": "parent", "type": "match", "value": parent}))
     else:
         must_list = _build_filters(components)
         # Restrict length by matching the END_MARK
-        must_list.append(FIELD_MATCH_VALUE %
-                         (_component_name(len(components)), END_MARK))
+        must_list.append(json.dumps(
+            {"field": _component_name(len(components)), "type": "match", "value": END_MARK}))
```
iksaif: could we do the json.dumps() were we join the constraints instead ? (`LUCENE_FILTER % ",\n            ".join(must_list)`)

biggraphite/drivers/lucene.py
```diff
@@ -72,7 +65,7 @@ def translate_to_lucene_filter(components):
         if gs_index == len(components) - 1:
             # No define length: match on components before the GLOBSTAR
             components.pop()
-            must_list = _build_filters(components)
+            lucene_filter["filter"]["must"] = _build_filters(components)
```
iksaif: add this to must_list, and do `lucene_filter['filer']['must']` later in the code to remove some duplication

Issue #296, opened 2017-10-03
bg-carbon-cache processes stop working


n-zeed:
Hello guys,

I am running BigGraphite 0.8.9 with following:

2 cache servers with 6 bg-carbon-cache processes, 1 relay process each
1 relay server with 1 relay processes
4 cassandra nodes (docker image) with replication factor 1 (tried with RF=3 got progation errors)

An another BigGraphite with 0.8.8 with:
4 cache servers with 6 bg-carbon-cache processes, 1 relay process each
1 relay server with 1 relay processes
12 cassandra nodes rpm install with replication factor 1 (tried with RF=2 got progation errors)

each cluster is receiving an aprox of 400K-500K metrics per minute and after a while bg-caches processes stop working, relays keep sendin data but daemons stop processing data (no updates.log) although the processes itself if up this is more common with 0.8.8 as soon as the first process stops the others follow.

in dev (0.8.9) I've had only 1 process per cache server stop, but still it stops for no apparent reason, as I said there are no exceptions on logs for graphite nor cassandra.

Any ideas on what could be going on?

iksaif:
Between 0.8.9 and 0.8.8 there should not be many differences appart form the new `created_at` field in `metrics_metadata`.

On top of updates.log carbon should log a few more things, could you redirect them either to syslog (`--syslog`), or to stdout (`--debug --nodaemon`).


For the propagation errors, things should work fine as long as tables are created in advance, that's quite annoying that Cassandra doesn't provide that out the box. I'll see what can be done to improve this, but it's hard without centralized locking .. maybe a tool that reads your carbon.conf and creates all the tables  on startup. It's weird because we neither really had issues with that.

n-zeed:
appreciate the offer and we would like that if you have a slack channel you could invite us, we can provide more details of the setup. We are running everything within docker but plan to test installing BG in a vm see if that helps with our problem.

n-zeed:
![carbon_stops_processing](https://user-images.githubusercontent.com/9345960/31198282-81ae2dc6-a908-11e7-8148-d3719c01e53f.PNG)

iksaif:
We can discuss on https://gitter.im/biggraphite/, I'll be there tomorrow from 9am to 6pm CET.

n-zeed:
[...]

PR #333, opened 2018-04-23
Add non-zero timeout to event.wait()

ofosos:
In response to #296


biggraphite/accessor.py
```diff
@@ -91,7 +91,7 @@ def on_done(exception):
         event.set()
 
     async_function(*args, on_done=on_done, **kwargs)
-    event.wait()
+    event.wait(3.0)
```
iksaif: can you add a code comment explaining why, with the link to the issue ?
Thanks !

ofosos:
Is this ok?

PR #340, opened 2018-07-03
ES: implement glob_metrics and glob_directories

Thib17:


biggraphite/drivers/elasticsearch.py
```diff
@@ -20,7 +20,9 @@
 import datetime
 import uuid
 import logging
+import six
 import elasticsearch
+from elasticsearch_dsl import Search
```
iksaif: just `import elasticsearch_dsl`

biggraphite/drivers/elasticsearch.py
```diff
@@ -286,20 +285,116 @@ def delete_directory(self, name):
         self._directory_names.remove(name)
 
     @staticmethod
```
iksaif: if this are static methods, moving them outside of the class might make them easier to read /test

biggraphite/drivers/elasticsearch.py
```diff
@@ -286,20 +285,116 @@ def delete_directory(self, name):
...
+        # TODO really implement it
+        if aggregate_dirs == True:
+            return []
+
+        if glob == "":
+            return []
+
+        search = Search().using(self.client).index(self._index).source('name')
+        # TODO Only if no globstars !!!
+        # en attendant, we assert no globstars
```
iksaif: for globstart we can probably use a simple regexp on name (which means special handling for AnySequence/AnyChar)

biggraphite/drivers/elasticsearch.py
```diff
@@ -286,20 +285,116 @@ def delete_directory(self, name):
...
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).glob_metric_names(glob)
-        return iter(self.__glob_names(self._metric_names, glob))
+        return iter(self.__glob_names(glob))
 
     def glob_directory_names(self, glob):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).glob_directory_names(glob)
-        return iter(self.__glob_names(self._directory_names, glob))
+        return iter(self.__glob_names(glob, aggregate_dirs=True))
```
iksaif: for elasticsearch, metrics and name will be very different, so using __glob_names will not make much sense.

biggraphite/drivers/elasticsearch.py
```diff
@@ -286,20 +285,116 @@ def delete_directory(self, name):
...
+
+        components = self.__glob_parser.parse(glob)
+
+        if self.__glob_parser.is_fully_defined(components):
+            search = search.filter('term',**{"name" : ".".join(self._components_from_name(glob))})
+        else:
+            for i, c in enumerate(components):
+                if len(c) == 1:
+                    type, value = self.__parse_simple_component(c)
+                else:
```
iksaif: `type` is a reserved keywoird

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: Ok. Still need to be implemented.

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
```diff
@@ -207,6 +209,80 @@ class InvalidArgumentError(Error, bg_accessor.InvalidArgumentError):
...
+        return 'regexp', '[^' + ''.join(value.values) + ']'
+    elif isinstance(value, bg_glob.CharIn):
+        return 'regexp', '[' + ''.join(value.values) + ']'
+    elif isinstance(value, bg_glob.SequenceIn):
+        return 'terms', value.values
+    elif isinstance(value, bg_glob.AnyChar):
+        return 'wildcard', '?'
+    else:
+        raise Error("Unhandled type '%s'" % value)
+
```
iksaif: can you start adding unit tests for these (it's easy as they don't depend on es).

if you want to implement them in a follow up review to avoid conflicts, add a TODO

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
+        # Handle all other use cases.
+        for i, c in enumerate(components):
+            if len(c) == 1:
+                filter_type, value = parse_simple_component(c)
+            else:
+                filter_type, value = parse_complex_component(c)
+
+            if filter_type:
+                search = search.filter(filter_type, **{"p%d.keyword" % i: value})
+        return False, search
```
iksaif: this can be unit tested too.

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
+
+        components = self.__glob_parser.parse(glob)
+
+        # Handle glob with globstar(s).
+        globstars = components.count(bg_glob.Globstar())
+        if globstars:
+            name_regexp = "\\.".join([_parse_regexp_component(c) for c in components])
+            return True, search.filter('regexp', **{"name": name_regexp})
+
+        # Handle fully defined glob.
```
iksaif: There is another optimization that can be done here when there is something that looks like  'a.b.c.*.*.*': we can do name: 'a.b.c.*' instead of searching individual components.

Add a TODO for this.

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
+        if glob == "":
+            return []
+
+        has_globstar, search = self._search_metrics_from_glob(glob)
+        if has_globstar:
+            # FIXME (t.chataigner) Add a log or raise exception.
+            return []
+
+        glob_depth = glob.count(".")
+        search = search.filter('range', depth={'gte': glob_depth + 1})
```
iksaif: maybe add a comment explaining why

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
     def glob_directory_names(self, glob):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).glob_directory_names(glob)
-        return iter(self.__glob_names(self._directory_names, glob))
+        if glob == "":
+            return []
+
+        has_globstar, search = self._search_metrics_from_glob(glob)
+        if has_globstar:
+            # FIXME (t.chataigner) Add a log or raise exception.
```
iksaif: change FIXME in TODO if you don't want to implement it now

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
+        glob_depth = glob.count(".")
+        search = search.filter('range', depth={'gte': glob_depth + 1})
+        search = search.extra(from_=0, size=0)  # Do not return metrics.
+
+        search.aggs.bucket('distinct_dirs', 'terms', field="p%d.keyword" % glob_depth)
+
+        log.debug(search.to_dict())
+        response = search.execute()
+
+        glob_base = glob.rsplit('.', 1)[0]
```
iksaif: maybe add a comment explaining that this might not be the exact behavior of other drivers.

biggraphite/drivers/elasticsearch.py
```diff
@@ -399,21 +472,78 @@ def delete_metric(self, name):
...
+        has_globstar, search = self._search_metrics_from_glob(glob)
+        if has_globstar:
+            search = search.filter('range', depth={'gte': glob.count(".")})
+        else:
+            search = search.filter('term', depth=glob.count("."))
+        search = search.extra(from_=0, size=MAX_QUERY_SIZE)
+
+        # TODO (t.chataigner) try to move the sort in the ES search.
+        log.debug(search.to_dict())
+        results = [h.name for h in search.execute()]
```
iksaif: is it possible to return a generator here instead of a list ? (I'm not sure what biggraphite expects)

biggraphite/drivers/elasticsearch.py
Thib17: It's already implemented, right ?

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: TODO added

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: TODO added

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: Done

biggraphite/drivers/elasticsearch.py
Thib17: Done

iksaif:
please rebase :)

PR #343, opened 2018-07-04
Rename length to depth to bring clarity

fedj:
Today, the length is not accurate as a.b would have a length of 1
However, renaming length to depth would be clearer (a.b would have a depth of 1)

### DIFF
--- a/biggraphite/drivers/elasticsearch.py
+++ b/biggraphite/drivers/elasticsearch.py
@@ -76,7 +76,7 @@ INDEX_SETTINGS = {
     "mappings": {
         "_doc": {
             "properties": {
-                "length": {"type": "long"},
+                "depth": {"type": "long"},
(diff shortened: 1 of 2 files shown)


adericbourg:
This may be nit-picking, anyhow, I am wondering if keeping the same value (_ie_ not +1-ing it) but renaming it to `depth` could be more relevant. 

I feel like:

* a depth of 0 clearly states that it is the root
* a depth of 1 states that is it the 1st level after the root
* a length of 1 is misleading (is it the number of components? the length of the metric path?...)

iksaif:
@Thib17 length or depth ?

Thib17:
I agree with @adericbourg and vote `depth`.

fedj:
Done

fedj:
Done

PR #345, opened 2018-07-05
ES: implement get_metric and has_metric

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -424,7 +424,28 @@ def get_metric(self, metric_name, touch=False):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).get_metric(metric_name, touch=touch)
         metric_name = ".".join(_components_from_name(metric_name))
-        return self._name_to_metric.get(metric_name)
+        metric = self._name_to_metric.get(metric_name)
```
iksaif: this part is copied from the "Memory" driver, it should be removed.

biggraphite/drivers/elasticsearch.py
```diff
@@ -424,7 +424,28 @@ def get_metric(self, metric_name, touch=False):
...
+
+        # Metric is not in the cache: search in ES
+        search = elasticsearch_dsl.Search()
+        search = search.using(self.client) \
+            .index("%s*" % self._index_prefix) \
+            .source(['name', 'config']) \
+            .filter('term', name=metric_name)
+        response = search.execute()
+
+        if response is None or response.hits.total is not 1:
```
iksaif: `is not` -> `!=`

biggraphite/drivers/elasticsearch.py
```diff
@@ -424,7 +424,30 @@ def get_metric(self, metric_name, touch=False):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).get_metric(metric_name, touch=touch)
```
iksaif: Add a `TODO: ` to handle `touch=True` correctly (see cassandra.py, it's the part that will update `updated_on`)

biggraphite/drivers/elasticsearch.py
```diff
@@ -424,7 +424,30 @@ def get_metric(self, metric_name, touch=False):
...
+
+        # Metric is not in the cache: search in ES
+        search = elasticsearch_dsl.Search()
+        search = search.using(self.client) \
+            .index("%s*" % self._index_prefix) \
+            .source(['config']) \
+            .filter('term', name=metric_name)
+        response = search.execute()
+
+        if response is None or response.hits.total is not 1:
```
iksaif: `is not` -> `!=`

biggraphite/drivers/elasticsearch.py
adericbourg: Done

biggraphite/drivers/elasticsearch.py
adericbourg: Done

biggraphite/drivers/elasticsearch.py
adericbourg: Done

biggraphite/drivers/elasticsearch.py
```diff
@@ -119,6 +121,12 @@
 DEFAULT_PORT = 9200
 DEFAULT_TIMEOUT = 10
 
+MINUTE = 60
```
iksaif: This seems copied from cassandra.py, could be worth putting them in some common files

biggraphite/drivers/elasticsearch.py
```diff
@@ -566,10 +576,31 @@ def has_metric(self, metric_name):
...
-        # FIXME This hack make it possible to use glob_metric_names with bgutil list
-        # This need to be really implemented.
-        return self.make_metric(metric_name, bg_accessor.MetricMetadata())
+
+        metric = self.__get_metric(metric_name)
+
+        if touch:
+            self.__touch_metadata_on_need(metric, metric.updated_on)
+
+        return self.make_metric(
```
iksaif: could be worth reusing the timestamps here: see https://github.com/criteo/biggraphite/blob/master/biggraphite/accessor.py#L647

biggraphite/drivers/elasticsearch.py
```diff
@@ -566,10 +576,31 @@ def has_metric(self, metric_name):
...
+
+    def __get_metric(self, metric_name):
+        search = elasticsearch_dsl.Search()
+        search = search.using(self.client) \
+            .index("%s*" % self._index_prefix) \
+            .source(['uuid', 'config', 'updated_on']) \
+            .filter('term', name=metric_name)
+        response = search.execute()
+
+        if response is None or response.hits.total != 1:
```
iksaif: we will end up having duplicates since there is no hard constraint against it

sort by updated_on and add a 'size: 1' limit.

adericbourg: Done

biggraphite/drivers/elasticsearch.py
```diff
@@ -601,10 +634,24 @@ def fetch_points(self, metric, time_start, time_end, stage, aggregated=True):
...
+        # TODO: state if we should move the document from its index to
+        # the current (today) index
+        data = {
+            "doc": {
+                "updated_on": datetime.datetime.now()
+            }
+        }
+        self.client.update(
+            index=index,
+            doc_type='_doc',
```
iksaif: INDEX_DOC_TYPE

biggraphite/drivers/elasticsearch.py
adericbourg: Done

biggraphite/drivers/elasticsearch.py
adericbourg: Done

PR #349, opened 2018-07-06
Add scripts to help running test services

adericbourg:
Includes:
* Cassandra
* Carbon
* Graphite Web

### DIFF
--- a/scripts/cassandra_env.sh
+++ b/scripts/cassandra_env.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+cd $(git rev-parse --show-toplevel)
+
+source scripts/env.sh
+
+export CASSANDRA_VERSION=3.11.2
+export CASSANDRA_HOME=$(pwd)/apache-cassandra-${CASSANDRA_VERSION}
+
+if [ ! -d "${CASSANDRA_HOME}" ]; then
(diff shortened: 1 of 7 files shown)


scripts/cassandra_env.sh
```diff
@@ -0,0 +1,11 @@
+cd $(git rev-parse --show-toplevel)
```
iksaif: add a shebang ?

And a top-file comment to explain what it does and how to use it. Also applies to other file.

scripts/run_graphite_web.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+cd $(git rev-parse --show-toplevel)
+source scripts/env.sh
+
+GRAPHITE_WEB_CONF="${BG_VENV}/lib/python2.7/site-packages/graphite/local_settings.py"
```
iksaif: `python2.7` depends on the default python binary, which might be python 3

scripts/cassandra_env.sh
adericbourg: Done.

scripts/run_graphite_web.sh
adericbourg: Done.
(but unsure that I did it in a correct way)

PR #350, opened 2018-07-09
ES: fix `list` for root directories

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -564,8 +564,12 @@ def glob_directory_names(self, glob):
         # It returns the glob with the list of possible last component for a directory.
         # It doesn't return the list of fully defined directory names.
         glob_base = glob.rsplit('.', 1)[0]
-        results = ["%s.%s" % (glob_base, b.key)
-                   for b in response.aggregations.distinct_dirs.buckets]
+
+        buckets = response.aggregations.distinct_dirs.buckets
+        if glob_depth == 0:
+            results = [b.key for b in buckets]
+        else:
```
Thib17: You can move glob_base definition (and the comment) inside the `else`

PR #351, opened 2018-07-09
Rename TestCaseWithAccessor to TestCaseWithCassandraAccessor

adericbourg:
This test-case base class is dedicated to Cassandra and would be
misleading when starting writing tests for Elasticsearch accessor.

### DIFF
--- a/biggraphite/test_utils.py
+++ b/biggraphite/test_utils.py
@@ -207,7 +207,7 @@ class TestCaseWithFakeAccessor(TestCaseWithTempDir):
 @unittest.skipUnless(
     HAS_CASSANDRA, "CASSANDRA_HOME must be set to a >=3.5 install",
 )
-class TestCaseWithAccessor(TestCaseWithTempDir):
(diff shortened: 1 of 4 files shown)


biggraphite/test_utils.py
```diff
@@ -207,7 +207,7 @@ def make_metric(self, name, metadata=None, **kwargs):
 @unittest.skipUnless(
     HAS_CASSANDRA, "CASSANDRA_HOME must be set to a >=3.5 install",
 )
-class TestCaseWithAccessor(TestCaseWithTempDir):
```
iksaif: Thinking about the future Cassandra + Elasticsearch driver, shouldn't we keep TestCaseWithAccessor.

Maybe the ES and Cassandra classes should not override TestCase's methods and thay should be called explicitly instead

adericbourg: Composition is usually more effective than inheritance, so I would agree with that.
I'll abandon this PR.

PR #363, opened 2018-07-11
ES: fix regexp in value sequence

adericbourg:
Example:

    foo.bar.{baz,qux*}

was not returning the correct result: a 'terms' query did not
match the regexp.

This case is now handled.


adericbourg:
@iksaif It is when there is no regexp wildcard in the terms.

PR #365, opened 2018-07-12
ES: raise error when accessing points from Elasticsearch accessor

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -302,6 +300,10 @@ def _get_depth_from_components(components):
     return len(components) - 1
 
 
+def _raise_unsupported():
+    raise NotImplementedError("Elasticsearch accessor does not support point operations")
```
iksaif: /write/data/

biggraphite/drivers/elasticsearch.py
```diff
@@ -432,34 +426,19 @@ def insert_points_async(self, metric, datapoints, on_done=None):
...
-        if on_done:
-            on_done(None)
+        _raise_unsupported()
 
     def drop_all_metrics(self, *args, **kwargs):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).drop_all_metrics(*args, **kwargs)
         # Drop indices.
-        self._metric_to_points.clear()
         self._name_to_metric.clear()
```
iksaif: _name_to_metric / _directory_names should be removed too, there are leftovers from MemoryAccessor

biggraphite/drivers/elasticsearch.py

biggraphite/drivers/elasticsearch.py

PR #369, opened 2018-07-13
Move make_metric to accessor class

fedj:
It was duplicated in both accessors


biggraphite/accessor.py
```diff
@@ -694,6 +702,8 @@ class Accessor(object):
 
     __metaclass__ = abc.ABCMeta
 
+    _UUID_NAMESPACE = uuid.UUID('{00000000-1111-2222-3333-444444444444}')
```
adericbourg: What about moving this to a static value (instead of an instance value)?

PR #376, opened 2018-07-17
Add unit tests on ES interactions

fedj:


tests/drivers/test_elasticsearch.py
```diff
@@ -256,6 +257,68 @@ def test_combination_of_glob_expressions_should_be_concatenated(self):
         self.assertEqual(result, "foo(bar|baz)[^ab].*")
 
 
+class ElasticSearchTestAccessor(unittest.TestCase):
+
+    def setUp(self):
+        self._accessor = bg_elasticsearch._ElasticSearchAccessor()
+        self._old_connect_method = self._accessor.connect
```
adericbourg: This is never used. What was your intent backing this?

Why is it different from `client`?

tests/drivers/test_elasticsearch.py
```diff
@@ -256,6 +257,68 @@ def test_combination_of_glob_expressions_should_be_concatenated(self):
...
+
+        meta_dict = {
+            "aggregator": Aggregator.last,
+            "retention": Retention.from_string("60*1s:60*60s"),
+            "carbon_xfilesfactor": 0.3,
+        }
+        metadata = MetricMetadata(**meta_dict)
+        self._metric = self._accessor.make_metric("a.b", metadata)
+
+        def response_mock(index, doc_type, body):
```
adericbourg: I'm wondering if you could do:

```
def response_mock(*unused):
```
to denote that the parameters are not used.

tests/drivers/test_elasticsearch.py
```diff
@@ -256,6 +257,68 @@ def test_combination_of_glob_expressions_should_be_concatenated(self):
...
+                    "read_on": None,
+                    "config": {
+                        "aggregator": "last",
+                        "retention": "60*1s:60*60s",
+                        "carbon_xfilesfactor": 0.3,
+                    }
+                }
+            }]}}
+
+        self._accessor.client.search = response_mock
```
adericbourg: What maked you handle differently this case from `exists` for instance?

In other words, what would happen with a 
```
self._accessor.client.search = mock.Mock(return_value= ...)
```

adericbourg:
@fedj What do you want to do with this PR?

PR #380, opened 2018-07-17
Recreate metric in current index if index is old.

fedj:
If the index of the fetched document is not the same as the one of today,
recreate the metric in the current index.

### DIFF
--- a/biggraphite/drivers/elasticsearch.py
+++ b/biggraphite/drivers/elasticsearch.py
@@ -31,7 +31,7 @@ from biggraphite import glob_utils as bg_glob
 from biggraphite.drivers import _utils
 
 from biggraphite.drivers.ttls import DEFAULT_READ_ON_TTL_SEC, DEFAULT_UPDATED_ON_TTL_SEC
-from biggraphite.drivers.ttls import str_to_datetime, str_to_timestamp
+from biggraphite.drivers.ttls import str_to_datetime, str_to_timestamp, datetime_to_str
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


biggraphite/drivers/elasticsearch.py
```diff
@@ -647,18 +650,30 @@ def touch_metric(self, metric_name):
...
+        document_id = document.uuid
+        updated_on = datetime.datetime.now()
         data = {
             "doc": {
-                "updated_on": datetime.datetime.now()
+                "updated_on": updated_on
             }
         }
         self.__update_document(data, index, document_id)
+        document.updated_on = updated_on.replace(microsecond=0).isoformat()
```
iksaif: can you add a code comment to explain `replace(microsecond=0).isoformat()` ? I though you could just put a datetime here.

biggraphite/drivers/elasticsearch.py
```diff
@@ -600,29 +600,32 @@ def get_metric(self, metric_name, touch=False):
...
+            document.config.to_dict()
         )
         return self.make_metric(
-            metric_name,
+            document.name,
             metadata,
-            created_on=str_to_datetime(metric.created_on),
-            updated_on=str_to_datetime(metric.updated_on),
-            read_on=str_to_datetime(metric.read_on)
+            created_on=str_to_datetime(document.created_on),
```
iksaif: I don't remember why these are strings, do you know ? can we put a comment explaining why ?

biggraphite/drivers/elasticsearch.py
```diff
@@ -699,15 +714,15 @@ def map(
         for i, metric in enumerate(metrics):
             callback(metric, i, total)
 
-    def __touch_metadata_on_need(self, metric):
-        if not metric.updated_on:
+    def __touch_metadata_on_need(self, document):
```
iksaif: What do you think about renaming this `__maybe_set_updated_on` and same for the `read_on` methods ? That would be less confusing (potentially, we could do that in cassandra.py too)

biggraphite/drivers/elasticsearch.py
fedj: I'll extract a function in utils then

biggraphite/drivers/elasticsearch.py
fedj: I don't, I'll take a look why

biggraphite/drivers/elasticsearch.py
fedj: Let's do this in another refactoring PR

biggraphite/drivers/elasticsearch.py
fedj: The search API of the dsl get strings. We could use other APIs from the dsl. Added a comment in the code

PR #381, opened 2018-07-18
ES: delete metric and delete directory

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -486,14 +486,29 @@ def update_metric(self, name, updated_metadata):
         self.create_metric(updated_metric)
 
     def delete_metric(self, name):
-        # TODO: Implement
-        log.warn("update_metric is not implemented")
-        pass
+        name = bg_accessor.sanitize_metric_name(name)
+
+        query = elasticsearch_dsl.Search() \
```
iksaif:
```
+        query = elasticsearch_dsl.Search() \
+            .using(self.client) \
+            .index("%s*" % self._index_prefix)
```

this seems duplicated, maybe we can add a helper to do that ?

adericbourg: Sure. This is used in other methods: do you agree if I do that in another PR?

iksaif: ok !

adericbourg: Done with #383

PR #384, opened 2018-07-18
Refactoring: remove touch attribute from get_metric

fedj:
A boolean modifying the behavior of a method can be confusing.
Callers with the need to touch the metric will have to call it manually

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -1364,15 +1364,6 @@ class _CassandraAccessor(bg_accessor.Accessor):
             self.__select_directory_statement, (encoded_directory, )))
         return bool(result)
 
-    def __touch_metadata_on_need(self, metric_name, updated_on):
-        if not updated_on:
-            delta = self.__metadata_touch_ttl_sec + 1
-        else:
-            delta = int(time.time()) - int(updated_on / 1000)
-
-        if delta >= self.__metadata_touch_ttl_sec:
-            self.touch_metric(metric_name)
-
     def __query_key(self, statement, params):
         if isinstance(statement, six.string_types):
             return statement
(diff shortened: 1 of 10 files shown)


biggraphite/drivers/cassandra.py
```diff
@@ -1397,21 +1390,8 @@ def get_metric(self, metric_name, touch=False):
         if not uid or not config:
             return None
 
-        if touch:
-            # Small trick here: we also check that the parent directory
-            # exists because that's what we check to create the directory
-            # hierarchy.
-            # We do it only on 'touch==True' because we know that the caller
-            # accepts a small performance penalty already.
```
iksaif: I think this part was actually useful because it allows us to remove some entropy by recreating directories when needed.

Can you either keep it and always do it ?

biggraphite/drivers/cassandra.py
```diff
@@ -1689,17 +1669,22 @@ def syncdb(self, retentions=None, dry_run=False):
...
             return
 
-        self._execute_async_metadata(
-            self.__touch_metrics_metadata_statement,
-            (metric_name,)
-        )
+        if not metric.updated_on:
+            delta = self.__metadata_touch_ttl_sec + 1
+        else:
+            delta = int(time.time()) - int(metric.updated_on / 1000)
```
iksaif: is updated_on a timestamp or a datetime ?

biggraphite/plugins/carbon.py
```diff
@@ -146,10 +146,11 @@ def write(self, metric_name, datapoints):
         metric_name = self.encode(metric_name)
 
         # Get a Metric object from metric name.
-        metric = self.cache.get_metric(metric_name=metric_name, touch=True)
+        metric = self.cache.get_metric(metric_name=metric_name)
         if not metric:
             raise Exception('Could not find %s' % (metric_name))
 
+        self.accessor.touch_metric(metric)
```
iksaif: Add a comment here explaining why

fedj: should we rename touch_metric then ?

iksaif: Not necessarily, just explaining that we call it here because we are writing and thus may want to update the timestamps might be enough. You can also just call touch_metric in insert_points()

biggraphite/drivers/cassandra.py
fedj: the same as in cassandra. Btw, it's just code that moved from touch_metadata_on_need to here (modulo the fact that it's now stored in metric object directly)

biggraphite/drivers/cassandra.py
fedj: Done

biggraphite/drivers/cassandra.py
iksaif: in cassandra it's a timeuuid :P , but ok if it just moved then it should be fine

iksaif:
can you rebase and merge ? thanks !

PR #386, opened 2018-07-18
ES: use query time range to filter metrics metadata

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -507,8 +507,8 @@ def delete_directory(self, name):
         query.delete()
 
     # TODO (t.chataigner) Add unittest.
-    def _search_metrics_from_components(self, glob, components):
-        search = self._create_search_query().source('name')
+    def _search_metrics_from_components(self, glob, components, start_time=None, end_time=None):
+        search = self._create_search_query(start_time, end_time).source('name')
```
iksaif: What do you think about giving a "Search" object to `_search_metrics_from_components` and letting the caller call `_create_search_query()` ? This way you would not have to give start_time/end_time to all these functions

biggraphite/drivers/elasticsearch.py
```diff
@@ -762,10 +769,15 @@ def __update_document(self, data, index, document_id):
...
         )
 
-    def _create_search_query(self):
-        return elasticsearch_dsl.Search() \
+    def _create_search_query(self, start_time=None, end_time=None):
+        search = elasticsearch_dsl.Search() \
             .using(self.client) \
             .index("%s*" % self._index_prefix)
+        if start_time is not None:
+            search = search.filter('range', updated_on={"gte": start_time})
```
iksaif: you need to add the touch_ttl potential delay here

biggraphite/drivers/elasticsearch.py

biggraphite/drivers/elasticsearch.py
adericbourg: Actual `updated_on` value should be greater or equal that the "real" one:

```
real_updated_on + touch_ttl >= updated_on >= real_updated_on
```
We want: 
```
real_updated_on >= start_time
```
so 
```
updated_on >= start_time
```
seems ok to me. 

We would get a bit more metrics than we should, at worse.

What would I miss here?

adericbourg: Oh, understood. 

I added a comment to explain in the code. Will be checked in tests in a coming commit for this PR.

tests/drivers/test_elasticsearch.py
```diff
@@ -344,6 +344,58 @@ def test_fetch_points_updates_read_on(self):
         # Try again, we read_on != None now
         self.accessor.fetch_points(metric, 0, 1, metric.retention[0])
 
+    def test_metric_out_of_search_bounds_should_not_be_found(self):
+        metric_name = "test_metric_out_of_search_bounds_should_not_be_found.a.b.c"
+        metric = self.make_metric(metric_name)
+        self.accessor.create_metric(metric)
+        self.accessor.flush()
+
+        now = datetime.datetime.now()
```
fedj: Maybe worth looking at metric.created_on instead. It will bring clarity and will allow to test strict boundaries (e.g. equal to the created_on)

tests/drivers/test_elasticsearch.py
```diff
@@ -344,6 +344,58 @@ def test_fetch_points_updates_read_on(self):
...
+            )
+            self.assert_iter_empty(should_not_be_found_metric)
+
+    def test_metric_in_search_bounds_should_be_found(self):
+        metric_name = "test_metric_in_search_bounds_should_be_found.a.b.c"
+        metric = self.make_metric(metric_name)
+        self.accessor.create_metric(metric)
+        self.accessor.flush()
+
+        now = datetime.datetime.now()
```
fedj: Same here.

tests/drivers/test_elasticsearch.py
```diff
@@ -344,6 +344,58 @@ def test_fetch_points_updates_read_on(self):
...
+            (now - 2 * one_week, now - one_week),
+            (now + one_week, now + 2 * one_week),
+            (None, now - one_week),
+            (now + one_week, None)
+        ]
+
+        for start_time, end_time in bounds:
+            # Without bounds, metric should be found (1 element in the iterator)
+            should_be_found_metrics = self.accessor.glob_metric_names(metric_name)
+            self.assert_iter_not_empty(should_be_found_metrics)
```
fedj: I understand why you do it but shouldn't it already be covered by other tests ? If you can remove this, you could factorize both unit tests

tests/drivers/test_elasticsearch.py
adericbourg: Done.

tests/drivers/test_elasticsearch.py
adericbourg: Done.

tests/drivers/test_elasticsearch.py
adericbourg: Done.

PR #394, opened 2018-07-24
improve tests and scripts

iksaif:
* improve scripts/
* remove all time.sleep() in tests and use freezegun instead


scripts/env.sh
```diff
@@ -35,3 +38,14 @@ fi
 
 # In test environment, limit to 12 components
 export BG_COMPONENTS_MAX_LEN=12
+
+export CASSANDRA_VERSION=3.11.2
+export ES_VERSION=6.3.1
+export ES_HOME=./.deps/ELASTICSEARCH_HOME
```
adericbourg: What should do this line?

PR #399, opened 2018-07-24
Mount tmpfs for Cassandra and Elasticsearch data

adericbourg:
If this works, it should improve test run time.

### DIFF
--- a/tools/mount-tmpfs.sh
+++ b/tools/mount-tmpfs.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -o xtrace
+set -o verbose
+
+# On Travis, /tmp is on tmpfs
+
+if [ -n "${CASSANDRA_VERSION}" ]; then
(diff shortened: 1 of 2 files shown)


tools/mount-tmpfs.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ -n "${CASSANDRA_VERSION}" ]; then
+    rm -rf .deps/apache-cassandra-${CASSANDRA_VERSION}/data
+    mkdir /dev/shm/cs-bg-data
+    ln -s /dev/shm/cs-bg-data .deps/apache-cassandra-${CASSANDRA_VERSION}/data
+fi
+
+if [ -n "${ES_VERSION}" ]; then
+    rm -rf .deps/elasticsearch-${ES_VERSION}/data
+    mkdir /dev/shm/es-bg-data
```
iksaif: you might have to use sudo here (and make sure sudo is available in travis)

adericbourg:
Does not work. Will retry (maybe) later.

PR #400, opened 2018-07-25
Mount tmpfs for Cassandra and Elasticsearch data

adericbourg:


iksaif:
Does it really work ? It doesn't seem to be faster than before.

adericbourg:
It does not seem to be faster, no :(
I guess this script can be useful as a local tool. I can remove it from the `.travis-ci.yml` file...

iksaif:
Yep, let's remove it, I think they have magic disks that are already pretty fast and the bottlenect might be CPU

PR #401, opened 2018-07-25
Install dependencies only when they are not present

adericbourg:


CONTRIBUTING.md
```diff
@@ -40,30 +40,10 @@ By convention:
...
-# Install the libffi-dev package from your distribution before running pip install
-pip install -r requirements.txt
-pip install -r tests-requirements.txt
-pip install -e .
-
-# Install Cassandra and Elasticsearch
-export CASSANDRA_VERSION=3.11.2
-export ES_VERSION=6.3.1
-export CASSANDRA_HOME=$(pwd)/apache-cassandra-${CASSANDRA_VERSION}
-./tools/install-deps.sh
```
iksaif: I kind of like the manual instructions because it explained how things work.. We can see if we re-add them later.

PR #411, opened 2018-07-25
Make es schema configurable

fedj:
Future work could override the default path if needed

### DIFF
--- a/biggraphite/drivers/elasticsearch.py
+++ b/biggraphite/drivers/elasticsearch.py
@@ -55,73 +55,6 @@ log = logging.getLogger(__name__)
 
 INDEX_DOC_TYPE = "_doc"
 
-# TODO: Make that configurable (in a file), this will be particularly important
-# for the number of shards and replicas.
-INDEX_SETTINGS = {
-    "settings": {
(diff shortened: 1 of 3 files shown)


biggraphite/drivers/elasticsearch.py
```diff
@@ -121,6 +54,8 @@
 DEFAULT_USERNAME = os.getenv("BG_ELASTICSEARCH_USERNAME")
 DEFAULT_PASSWORD = os.getenv("BG_ELASTICSEARCH_PASSWORD")
 
+ES_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), '../../share/es_schema')
```
iksaif: es_schema.json ?

biggraphite/drivers/elasticsearch.py
```diff
@@ -378,6 +314,9 @@ def _connect(self):
         log.info("Connected: %s" % es.info())
         self.client = es
 
+        with open(ES_SCHEMA_PATH, "r") as es_schema_file:
```
iksaif: make the path configurable ?

PR #417, opened 2018-07-30
Move test_* files to tests folder

fedj:


benchmarks/accessor_bench.py
```diff
@@ -7,7 +7,7 @@
 import six
 
 from biggraphite import accessor as bg_accessor
-from biggraphite import test_utils as bg_test_utils
+from tests import test_utils as bg_test_utils
```
adericbourg: You could re-order the imports here

PR #420, opened 2018-08-01
Improve documentation

adericbourg:
* Remove Gemnasium (service is closed)
* Add Elasticsearch as supported metadata backend
* Document scripts to use tmpfs for tests
* Typo


USAGE.md
```diff
@@ -36,16 +37,58 @@ You'll need to start `bg-carbon-cache` instead of `carbon-cache` to run carbon w
...
 BG_CASSANDRA_KEYSPACE = 'biggraphite'
 BG_CASSANDRA_CONTACT_POINTS = '127.0.0.1'
 ```
 
 You can optionally keep the StandardFinder to enable double reads from whisper
-and biggraphite.
\ No newline at end of file
+and biggraphite.
+
+To use Elasticsearch as a metadata backend, replace `BG_DRIVER` parameter by
```
fedj: The following seems copy/pasted. Can we add a link instead ?

adericbourg: I'll simplify this, then

PR #424, opened 2018-08-01
Transform DRIVERS to a dict

fedj:


Thib17:
What does it fix ?
Isn't that saffer to keep it in a frozenset ?

fedj:
A dict is (IMO) way easier to understand than a set of pairs (which basically is the same concept). frozenset is immutable yes but after the refactorings of yesterday, it can now be private to this module.

Thib17:
afaik nothing is really private in python. It's more like a writting convention.

iksaif:
Can you rebase and fix the tests ?

fedj:
Actually, the API of dict changed between python 2.7 and 3 so I'll cancel this one

PR #430, opened 2018-08-03
bump to 0.13.2

Thib17:


CHANGELOG.md
```diff
@@ -6,7 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [0.13.1] - 2018-05-02
+## [0.13.2] - 2018-08-03
+
+### removed
```
adericbourg: For consistency with the rest of the file, that would be _Removed_ (with an upper case)

adericbourg: For consistency with the rest of the file, that would be _Removed_ (with an upper case)

Thib17: Done

PR #437, opened 2018-08-08
Fix metadata read parsing in Cassandra driver.

Thib17:
We start getting AttributeError for metric whose metadata cache
has expired, and we explictly read from Cassandra.
Example error log:
```
07/08/2018 17:53:21 :: [console] Error writing to carbon.agents.tchataigner-laptop-a.process_virtual_memory_bytes: 'int' object has no attribute 'timetuple'```

### CODE REVIEW

File: biggraphite/drivers/cassandra.py [hunk 1]
```diff
@@ -1387,7 +1387,7 @@ def get_metric(self, metric_name):
             return None
         uid = result[0]
         config = result[1]
-        updated_on = result[2]
+        updated_on = datetime.datetime.fromtimestamp(result[2]/1000)
```
adericbourg: Cannot we reproduce the issue with a test and only then fix it?

Thib17: Done

PR #445, opened 2018-08-16
Implement stats, map and clean for elasticsearch

geobeau:


biggraphite/cli/command_stats.py
```diff
@@ -115,35 +113,31 @@ def run(self, accessor, opts):
...
         See command.CommandBase.
         """
-        self.ns = Namespaces(opts.conf)
         accessor.connect()
-        accessor.map(
-            self.stats,
-            start_key=opts.start_key, end_key=opts.end_key,
-            shard=opts.shard, nshards=opts.nshards
-        )
+        if accessor.get_accessor_name() == "elasticsearch":
```
Thib17: You could use opts.driver or opts.metadata_driver

biggraphite/drivers/cassandra.py
```diff
@@ -1694,12 +1698,19 @@ def touch_metric(self, metric):
                 (metric.name,)
             )
 
-    def map(self, callback, start_key=None, end_key=None, shard=1, nshards=0, errback=None):
+    def map(self, namespaces, start_key=None, end_key=None, shard=1, nshards=0, errback=None):
```
Thib17: This is not a map function anymore. Could you create a stats() function instead that call map() with the right callback. And make the command_stats call stats() only.

biggraphite/accessor.py
```diff
@@ -315,26 +320,8 @@ def insert_downsampled_points_async(self, metric, datapoints, on_done=None):
...
-          start_key: string, start at key >= start_key.
-          end_key: string, stop at key < end_key.
-          shard: int, shard to repair.
-          nshards: int, number of shards.
-
-        """
-        assert shard >= 0
-        assert nshards > 0
-        assert shard < nshards
+    def map(self, namespaces, **kwargs):
```
Thib17: keep the idiomatic map function here. If it's not generic, we may move it to cassandra specific.

biggraphite/drivers/cassandra.py
geobeau: Done

biggraphite/accessor.py
geobeau: Done

biggraphite/cli/command_stats.py
geobeau: No, it's not really reliable. I could reimplement a function based on accessor_from_settings but I'm not sure it's simplier

biggraphite/drivers/hybrid.py
```diff
@@ -108,10 +112,10 @@ def insert_downsampled_points_async(self, metric, datapoints, on_done=None):
         # Updating metadata here would be too expensive
         self._data_accessor.insert_downsampled_points_async(metric, datapoints, on_done)
 
-    def map(self, callback, start_key=None, end_key=None, shard=0, nshards=1, errback=None):
```
iksaif: You probably want to keep "map" since it will get implemented at some point. Just add stats()

biggraphite/drivers/memory.py
```diff
@@ -67,6 +67,10 @@ def _components_from_name(metric_name):
     def _metric_names(self):
         return self._name_to_metric.keys()
 
+    def get_accessor_name(self):
```
iksaif: I would put that as a static member instead, but that's a question of taste

biggraphite/drivers/elasticsearch.py
```diff
@@ -664,39 +660,129 @@ def _callback(m, i, t):
...
+        index_suffix = self._index_suffix
+
+        indices_datetime = []
+        # TODO, we could also do this:
+        # self.client.cat.indices(
+        #   index="%s*" % self._index_prefix,
+        #   format="json", h="index,creation.date"
+        # )
+        for index in self.client.indices.get("%s*" % self._index_prefix):
+            # Hack if we use week as strptime cannot parse without weekday
```
Thib17: Is this hack really necessary ? Isn't parsedatetime able to manage it ?

biggraphite/cli/command_stats.py
```diff
@@ -117,13 +120,21 @@ def run(self, accessor, opts):
...
-            end_key=opts.end_key,
-            shard=opts.shard,
-            nshards=opts.nshards,
-        )
+
+        if accessor.TYPE.startswith("elasticsearch+"):
+            accessor = accessor._metadata_accessor
+
+        if accessor.TYPE == "elasticsearch":
+            # Elasticsearch has a better implementation.
```
Thib17: Can't we call a generic stats function, that in cassandra implementation would call map() as needed ?

iksaif: No, because in most implementations we iterate on metrics. For ES we iterate on patterns.

biggraphite/drivers/elasticsearch.py
iksaif: parsedatetime doesn't work well with week numbers :/

Thib17: ok

biggraphite/cli/command_stats.py
Thib17: ok

PR #449, opened 2018-08-21
bgutil web: add buttons to trigger clean/repair/stats

iksaif:


biggraphite/drivers/elasticsearch.py
```diff
@@ -617,7 +617,7 @@ def fetch_points(self, metric, time_start, time_end, stage, aggregated=True):
     def touch_metric(self, metric):
         """See the real Accessor for a description."""
         super(_ElasticSearchAccessor, self).touch_metric(metric)
-
+        self.__updated_on_ttl_sec = 5
```
Thib17: Should not be here, right ?

iksaif: woops

PR #452, opened 2018-08-22
elasticsearch: fix read_on, again

iksaif:
The previous code would update the latest index only, even
if the metric is on another one. Now we fetch the latest document
and we update read_on on this one.


fedj:
Do you think we can add a test ?

PR #458, opened 2018-08-28
Add `enum-compat` requirement

adericbourg:
Used in metric.py for `@enum.unique`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,7 @@ elasticsearch-dsl
 certifi
 
 # Core
+enum-compat
 six
 cachetools
 prometheus_client


requirements.txt
```diff
@@ -22,6 +22,7 @@ elasticsearch-dsl
 certifi
 
 # Core
+enum-compat
```
iksaif: nop, that breaks python3 .. the requirements has been added ton setup.py, you should not have to put it in requirements.txt

adericbourg: Ok, then either `CONTRIBUTING.md` either `scripts/env.sh` are not up to date. I had to install `enum34` manually.

iksaif: `pip install -e .` should install it, can you check ?

adericbourg: Running it manually works. 
I'll check what happened...

PR #459, opened 2018-08-29
[WIP] Change glob_metrics() to return metric metadata

adericbourg:


biggraphite/drivers/elasticsearch.py
```diff
@@ -536,8 +534,18 @@ def glob_metric_names(self, glob, start_time=None, end_time=None):
 
         # TODO (t.chataigner) try to move the sort in the ES search and return a generator.
         log.debug(json.dumps(search.to_dict(), default=str))
-        results = [h.name for h in search.execute()]
-        results.sort()
+
+        results = [self._document_to_metric(document) for document in search.execute()]
```
iksaif: How does it work when you have two indexes with the same document ? Which one will be returned ?

biggraphite/glob_utils.py
```diff
@@ -254,6 +255,40 @@ def maybe_matched_prefilter(metric):
     return list(filter(maybe_matched_prefilter, metric_names))
 
 
+class GlobMetricResult:
```
iksaif: I don't know if this intermediate class is really necessary .. but worst case it can't really hurt.

biggraphite/drivers/cassandra.py
```diff
@@ -1485,6 +1485,11 @@ def glob_directory_names(self, glob, start_time=None, end_time=None):
         directory_names = self.__glob_names("directories", glob)
         return bg_glob.filter_from_glob(directory_names, glob)
 
+    def glob_metrics(self, glob, start_time=None, end_time=None):
+        """Return a sorted list of metrics matching this glob."""
+        return [self.get_metric(metric_name)
```
iksaif: maybe it would be worth trying to use execute_concurrent here to have all the requests be started in parallel

biggraphite/drivers/elasticsearch.py
```diff
@@ -536,8 +534,32 @@ def glob_metric_names(self, glob, start_time=None, end_time=None):
...
+        results.sort(key=lambda metric: metric.name)
+
+        return results
+
+    @staticmethod
+    def _deduplicate_documents(documents):
+        """Deduplicate documents with same name over several indices to keep most recent one."""
+        documents_by_name = dict()
+        docs = list(documents)
+        docs.sort(key=lambda doc: doc.updated_on, reverse=True)
```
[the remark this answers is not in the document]
adericbourg: `updated_on` `None` values would be the last in the list (it is the "smallest" value).

`updated_on` is supposed to be always set, so `None` values would be broken documents that (I guess) can be filtered out. It does not seem to me to be a problem to have them last.

biggraphite/glob_utils.py
adericbourg: I wanted to avoid returning too complex anonymous types (tuple of lists of key-value pairs). This sounded more readable to me.

biggraphite/drivers/elasticsearch.py
```diff
@@ -535,11 +535,25 @@ def glob_metrics(self, glob, start_time=None, end_time=None):
...
+
+        for document in docs:
+            if document.name not in documents_by_name:
+                documents_by_name[document.name] = document
```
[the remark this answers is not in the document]
adericbourg: Done

biggraphite/drivers/cassandra.py
adericbourg: Done (that changed a few things in the accessor)

biggraphite/drivers/cassandra.py
```diff
@@ -1456,17 +1456,37 @@ def __query_key(self, statement, params):
...
+    def _get_metrics(self, metric_names):
+        metric_names = [".".join(self._components_from_name(metric_name)[:-1])
+                        for metric_name in metric_names]
+        metric_names = [bg_metric.encode_metric_name(metric_name) for metric_name in metric_names]
+
+        statements_with_params = [self._make_select_metric_statement(metric_name)
+                                  for metric_name in metric_names]
+
+        results = self._select_metrics(statements_with_params)
+        rows = [result[0] for (success, result) in results if success and result[0] is not None]
```
iksaif: might be worth not putting that as a list comprehension, I think it would be more readable..
In the same for loop you could do this and the _bind_metric(), this would have the advantage of starting bind_metric() earlier (which is good because it also calls Cassandra with a self.has_directory())

Also I don't know if an exception should be sent when success is False here.

biggraphite/drivers/elasticsearch.py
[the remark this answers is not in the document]
iksaif: ok ! :+1:

biggraphite/glob_utils.py
iksaif: ok :+1:

biggraphite/drivers/cassandra.py
adericbourg: * for comprehension: done
* about the exception, I'm wondering the same thing. I usually prefer returning no result rather that returning a wrong result. Here, the result would not be wrong but incomplete. Let's throw an exception and see if that is a problem when it (eventually) happens?

adericbourg:
:ok_hand:

geobeau:
Did you perform the test with success?

adericbourg:
@geobeau Yes, we ran some locally without noticeable issue.

iksaif:
After deployment can you post a latency graph here ? I'm curious :)

adericbourg:
@iksaif If it does not break the graph, sure! :)

PR #460, opened 2018-08-30
WIP Add tracing capabilities to biggraphite

geobeau:
POC for tracing with opencensus. 
Config will be done from django conf, without the middleware tracing will be mostly invisible (except the wrapping).
Waiting for some patch on the opencensus repository before merging

### DIFF
--- a/biggraphite/drivers/_utils.py
+++ b/biggraphite/drivers/_utils.py
@@ -18,6 +18,8 @@ from __future__ import print_function
 
 import threading
 
+from opencensus.trace import execution_context
+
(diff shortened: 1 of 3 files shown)


biggraphite/drivers/_utils.py
```diff
@@ -18,6 +18,8 @@
 
 import threading
 
+from opencensus.trace import execution_context
```
iksaif: is there a way to make it optional if the lib is not installed ?

geobeau: Done

PR #462, opened 2018-09-04
Add a dummy cache when no policy is required

adericbourg:


biggraphite/cache_factory.py
```diff
@@ -18,7 +18,11 @@
 
 
 CACHES = frozenset(
-    [("disk", metadata_cache.DiskCache), ("memory", metadata_cache.MemoryCache)]
+    [
+        ("disk", metadata_cache.DiskCache),
+        ("memory", metadata_cache.MemoryCache),
+        ("none", metadata_cache.NoneCache)
```
Thib17: Why not using just 'None' ? And then check `if cache` when using it

adericbourg: because it could be used in several places, so that would make the usage more complex. 

having this makes it transparent and seems less error-prone to me

PR #466, opened 2018-09-14
Hybrid accessor: no longer make calls to data accessor for metadata actions

adriengentil:
When using the hybrid mode, data and metadata accessors are both updating read_on and update_on metadata.

This is a problem because this makes mandatory to specify the metadata storage configuration for the data accessor while it is not used.

The solution consist in passing a flag preventing the data accessor to use the metadata storage when the hybrid mode is used.

### DIFF
--- a/biggraphite/accessor_factory.py
+++ b/biggraphite/accessor_factory.py
@@ -88,6 +88,14 @@ def accessor_from_settings(settings):
                 "Data driver is not provided. Please specify --data_driver"
             )
 
+        # The data accessor may still update the metadata in its own storage
+        # (update_on on insert_points(), read_on on fetch_points()). Since we
+        # are in hybrid mode, the metadata accessor is responsible for these
+        # actions, therefore we can disable the metadata handling in the data
+        # accessor.
+        enable_metadata_key = data_driver + '_enable_metadata'
+        settings[enable_metadata_key] = False
+
         metadata_accessor = _build_simple_accessor(metadata_driver, settings)
         data_accessor = _build_simple_accessor(data_driver, settings)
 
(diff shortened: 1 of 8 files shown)


biggraphite/accessor_factory.py
```diff
@@ -88,6 +88,10 @@ def accessor_from_settings(settings):
                 "Data driver is not provided. Please specify --data_driver"
             )
 
+        # disable metadata lookups in data driver
```
iksaif: Explain why here (and why it's ok to do so without breaking updated_on/read_on)

biggraphite/drivers/cassandra.py
```diff
@@ -964,25 +944,60 @@ def __init__(
...
+        self._meta_read_consistency = None
+        self._meta_serial_consistency = None
+        self._meta_background_consistency = None
+        self.__cluster_metadata = None  # setup by connect()
+        self.__session_metadata = None  # setup by connect()
+        self.__select_metric_metadata_statement = None  # setup by connect()
+        self.__update_metric_metadata_statement = None  # setup by connect()
+        self.__touch_metrics_metadata_statement = None  # setup by connect()
+        self.__metadata_touch_ttl_sec = None
+        if enable_metadata:
```
iksaif: Even when disabled it could be worth setting most of the attributes (as long as nothing actually connects to cassandra)

biggraphite/drivers/cassandra.py
```diff
@@ -964,25 +944,60 @@ def __init__(
...
+                meta_background_consistency
+            ]
 
     def connect(self):
         """See bg_accessor.Accessor."""
         super(_CassandraAccessor, self).connect()
         if not self.is_connected:
             self._connect_clusters()
-            self._prepare_statements()
+            if self._enable_metadata:
```
iksaif: the check could be in the function itself (and have an early return if True)

PR #470, opened 2018-10-08
WIP Implement a new way to do disk cache

geobeau:
Attempt at improving disk cache and simplify the implementation.

### DIFF
--- a/biggraphite/metadata_cache.py
+++ b/biggraphite/metadata_cache.py
@@ -30,16 +30,14 @@ import abc
 import logging
 import os
 import random
-import sys
 import threading
 import time
 import uuid
 from os import path as os_path
 
 import cachetools
-import lmdb
 import prometheus_client
-import six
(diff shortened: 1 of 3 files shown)


iksaif:
Note: if we finish re-writing the way metrics are globbed, the disk cache can go away because graphite-web will not need it anymore :)

geobeau:
It's used by carbon. But we are investigating if we need something like this or not

iksaif:
Carbon is not supposed to use the disk cache

geobeau:
Because it's too slow? Goal would be to have a cache that survives the restart of a process

iksaif:
Because it was too problematic (because of locking) and too slow (including the `has` function). The workaround was to make startup work without the cache and slowly create the metrics if necessary after startup (the rate is not configurable but could be, it also probably lacks back-pressure).

geobeau:
Ok, thanks for the informations!

PR #473, opened 2018-10-10
Improve observability on metric creation

adriengentil:
The metric creation process is asynchronous, metrics scheduled to be created are added to a queue and a thread process it to effectively create them in database.

This PR aims to monitor the enqueuing rate. The dequeuing rate where already monitored via `bg_creates` which I renamed `bg_creates_dequeued`.

### DIFF
--- a/biggraphite/plugins/carbon.py
+++ b/biggraphite/plugins/carbon.py
@@ -56,7 +56,13 @@ EXISTS_TIME = prometheus_client.Summary(
     "bg_exists_latency_seconds", "create latency in seconds"
 )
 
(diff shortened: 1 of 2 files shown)


biggraphite/plugins/carbon.py
```diff
@@ -315,7 +322,7 @@ def _createOneMetric(self):
 
         self.cache.create_metric(metric)
         self.tag(metric_name)
-        CREATES.inc()
+        CREATES_DEQUEUED.inc()
```
iksaif: I think you should move that line 307

adriengentil: Yes, I messed up. I will do a corrective PR.

PR #475, opened 2018-10-11
Add tracing capabilities for Biggraphite

geobeau:


biggraphite/drivers/_utils.py
```diff
@@ -104,3 +110,17 @@ def bool_from_str(value):
...
+def trace_accessor_func(func):
+    """Decorator for tracing of functions."""
+    if not execution_context:
+        return func
+
+    def tracer(self, *args, **kwargs):
+        if not hasattr(self, 'module_name'):
+            self.module_name = func.__module__.split('.')[-1]
+        tracer = execution_context.get_opencensus_tracer()
+        with tracer.span(name=self.module_name + '.' + func.__name__):
```
adericbourg: What made you chose string concatenation instead of string formatter?

biggraphite/drivers/cassandra.py
```diff
@@ -1160,6 +1160,7 @@ def __init__(
 
         self.metadata_enabled = enable_metadata
 
+    @_utils.trace_accessor_func
```
adericbourg: The name of the decorator does not seem explicit to me.

* `_utils` is very generic
* `_utils` stands for a private package / usage, it does not seem to be the intent here (can be used widely in the project)
* the function name could be imperative

My suggestion (feel free to suggest something else):
```
@tracing.trace
```

biggraphite/drivers/_utils.py
```diff
@@ -104,3 +110,17 @@ def bool_from_str(value):
         return value
 
     return str(value)
+
+
+def trace_accessor_func(func):
```
adericbourg: How could you unit test this function?

setup.py
```diff
@@ -41,7 +41,7 @@ def _read_reqs(relpath):
 
 
 if not PY3:
-  _INSTALL_REQUIRES += ('futures', 'enum34')
+    _INSTALL_REQUIRES += ('futures', 'enum34')
```
adericbourg: Looks like a rogue change :)

tests-requirements.txt
```diff
@@ -32,6 +32,9 @@ whisper
 # We could use Django<2.0 but we match what graphite requires instead.
 Django>=1.8,<1.11.99
 
+# Tracing
+opencensus
```
adericbourg: I'd like to understand your plans.

Currently:

* it makes tracing disabled in "real live" (dev machine, staging, prod)
* it enables tracing in tests

Right?

Further in time, how to you plan to be able to enable / disable tracing?

tests/drivers/test_tracing.py
```diff
@@ -0,0 +1,53 @@
...
+from opencensus.trace import tracer
+from opencensus.trace import execution_context
+
+from biggraphite.drivers import tracing
+
+
+class TestTracingTrace(unittest.TestCase):
+    """Test tracing integration of Biggraphite."""
+
+    def testDecoratorTrace(self):
```
adericbourg: Python function are usually `lower_snake_camel_case`

I'd also prefer the function name to explain what the test does (in a test run log, we have the function name but not the docstring). Example (feel free to find a better one): 

    test_decorating_with_tracing_function_should_create_a_span_with_same_name_as_foo

biggraphite/drivers/tracing.py
```diff
@@ -0,0 +1,34 @@
...
+    if not execution_context:
+        return func
+
+    def tracer(self, *args, **kwargs):
+        if not hasattr(self, 'module_name'):
+            self.module_name = func.__module__.split('.')[-1]
+        _tracer = execution_context.get_opencensus_tracer()
+        with _tracer.span(name="%s.%s" % (self.module_name, func.__name__)):
+            return func(self, *args, **kwargs)
+    return tracer
```
adericbourg: Does it have to be restricted to `biggraphite/drivers`?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+- enable tracing of biggraphite
```
adericbourg: Nitpicking: actually, this is not enabled. I'd rather say that it is _supported_.

setup.py
geobeau: Pylama spotted it so I thought I would fix it on the way

tests-requirements.txt
geobeau: To tracer you need to create a tracer somewhere. It can be done in different way but you can start it from the middleware of django. So if you install opencensus and integrate it in django (graphite web) tracing will be enabled. I will PR documentation on how to use it in a next PR.

biggraphite/drivers/tracing.py
geobeau: No, done

PR #482, opened 2018-11-15
Expose metrics from the elasticsearch driver

adriengentil:
Expose few metrics from the elasticsearch driver.


biggraphite/drivers/elasticsearch.py
```diff
@@ -646,6 +705,7 @@ def _document_to_metric(self, document):
             read_on=ttls.str_to_datetime(document.read_on),
         )
 
+    @GET_METRIC_LATENCY.time()
```
adericbourg: Why here and not on `get_metric`?

adriengentil: __get_document is called by get_metric, __touch_document and __update_read_on. I wanted to catch all calls that read a metric.

biggraphite/drivers/elasticsearch.py
```diff
@@ -36,11 +36,62 @@
...
+GLOB_METRICS_LATENCY = prometheus_client.Summary(
+    "bg_elasticsearch_glob_metrics_latency_seconds",
+    "glob metrics latency in seconds"
+)
+METRIC_STATS_LATENCY = prometheus_client.Summary(
+    "bg_elasticsearch_metric_stats_latency_seconds",
+    "metric stats latency in seconds"
+)
+READ_METRIC_LATENCY = prometheus_client.Summary(
+    "bg_elasticsearch_get_metric_latency_seconds",
```
adericbourg: You renamed the variable but not the actual name

adriengentil: Sorry, it should be good now.

adericbourg:
:+1:

PR #484, opened 2018-11-28
Share MetricMetadata objects as much as possible

adriengentil:
Today for every Metric object, we create a MetricMetadata object. This is a waste of memory since the same metadata can be shared by several metrics.

This change aims to rationalize the number of MetricMetadata objects created in biggraphite. I replaced all the calls to MetricMetadata's constructor by a call to MetricMetadata.create() who will create a MetricMetadata if necessary or return an existing one.

What do you think?

TODO:
- I know there is a problem with the xfilesfactor that is used as a dictionary key. It could be a problem because of the float representation.
- Add documentation/comments
- cleanup asserts in MetricMetadata.create()

### DIFF
--- a/tests/test_metric.py
+++ b/tests/test_metric.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import unittest
+import gc
+import prometheus_client
+import weakref
 
 from biggraphite import metric as bg_metric
 from tests import test_utils as bg_test_utils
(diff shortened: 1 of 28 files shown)


adriengentil:
I could make the items expire after a amount of time in the dict but the object will still exists if it is hold by another object (e.g: Metric or Downsampler). It could make things even worse as several instances of the same metadata could live within the program and wouldn't be reported by the size of the dictionary.

What could be done is to use weak references into the dict and clean it up from time to time, it look like  possible with https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary. I need to investigate a little more.

About the monitoring, you are right, I will add add a gauge to monitor the number of entries.

tests/test_metric.py
```diff
@@ -46,3 +49,87 @@ def test_sanitize_metric_name_should_handle_empty_value(self):
         self.assertEqual("", bg_metric.sanitize_metric_name(""))
         self.assertEqual("", bg_metric.sanitize_metric_name("."))
         self.assertEqual("", bg_metric.sanitize_metric_name(".."))
+
+
+class TestMetricMetadata(unittest.TestCase):
```
adericbourg: I feel like naming tests describing the expected behavior provides more info about the test, with a "thing SHOULD do stuff WHEN something". Eg:

* `test_create_should_define_default_values`
* `test_metadata_should_be_unique_for_a_given_parameter_tuple`
* `test_xfilesfactor_should_have_no_impact_on_metadata_uniqueness`
* ...

What is your thought about it?

tests/test_metric.py
```diff
@@ -46,3 +49,87 @@ def test_sanitize_metric_name_should_handle_empty_value(self):
...
+        retention1 = bg_metric.Retention.from_string("20*15s:1440*60s:48*3600s")
+        metadata1 = bg_metric.MetricMetadata.create(bg_metric.Aggregator.average, retention1, 0.1)
+        retention2 = bg_metric.Retention.from_string("20*15s:1440*60s:48*3600s")
+        metadata2 = bg_metric.MetricMetadata.create(bg_metric.Aggregator.average, retention2, 0.1)
+        self.assertIs(metadata1, metadata2)
+
+        # the returned instance should't depend on the order of the parameters
+        metadata3 = bg_metric.MetricMetadata.create(
+                                                    carbon_xfilesfactor=0.1,
+                                                    aggregator=bg_metric.Aggregator.average,
```
adericbourg: What about extracting variables / constants to show that this is actually the same values than for `metadata1`?

tests/test_metric.py
```diff
@@ -46,3 +49,87 @@ def test_sanitize_metric_name_should_handle_empty_value(self):
...
+                                                    retention=retention1)
+        self.assertIs(metadata1, metadata3)
+
+    def test_no_xfactor_uniqueness(self):
+        """Check that carbon_xfilesfactor has no influence on uniqueness."""
+        retention1 = bg_metric.Retention.from_string("20*15s:1440*60s:48*3600s")
+        metadata1 = bg_metric.MetricMetadata.create(bg_metric.Aggregator.average, retention1, 0.2)
+
+        retention2 = bg_metric.Retention.from_string("20*15s:1440*60s:48*3600s")
```
adericbourg: Same as above: I do not find obvious that you use the same retention for both metadata by reading the test (I have to check every character in the strings). 

What about sharing the value?

(and same for all the identical retention strings behind)

iksaif:
I'm curious, would you be able to share a graph of the effect of this change ?

PR #491, opened 2018-12-13
Adapt biggraphite to use the lastest version of prometheus client and bump to 0.13.12

adriengentil:
The Prometheus client changed the metric API from version 0.4.0.

This change makes it compatible to the newer API and freeze prometheus_client to version 0.5.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ certifi
 # Core
 six
 cachetools
-prometheus_client
+prometheus_client==0.5.0
 
 # Metadata cache
 lmdb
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -24,7 +24,7 @@ certifi
 # Core
 six
 cachetools
-prometheus_client
+prometheus_client==0.5.0
```
adericbourg: There should not be API changes between 0.5.* versions

What about using a range?

```
prometheus_client==0.5.*
```

PR #493, opened 2018-12-14
Some bgutils commands were print python objects instead of string representation

adriengentil:
For example with bgutil, the display is broken.
```
$ bg/bin/bgutil --loglevel=DEBUG --storage_dir=/opt/graphite/storage --driver=cassandra  --cassandra_contact_points=... --cassandra_contact_points_metadata=... --cassandra_concurrent_connections=50 list --graphite <query>
2018-12-14 16:34:40,909 [DEBUG] biggraphite.drivers.cassandra: [<biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1edf90>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1ede90>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1edd50>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1eddd0>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1edd10>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1edc50>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1edcd0>, <biggraphite.drivers.cassandra._CassandraExecutionRequest object at 0x7f1e8c1eda50>]
d <biggraphite.glob_utils.GlobDirectoryResult instance at 0x7f1e8de7c8c0>
```

I implemented __str__ on _CassandraExecutionRequest and GlobDirectoryResult in order to get the object's string.

I also removed the variable GlobDirectoryResult.value which seem to not be used in BG.


biggraphite/glob_utils.py
```diff
@@ -281,13 +281,15 @@ class GlobDirectoryResult:
     def __init__(self, name):
         """Create new instance of GlobMetricResult."""
         self.name = name
-        self.value = None
```
adericbourg: `GlobDirectoryResult` and `GlobMetricResult` are meant to be used indifferently and thus should have the same API. Removing it could lead to unexpected `AttributeError`. 

Maybe we could make it more clear / obvious... as there is no typing opportunity for now.

adriengentil: ok that explains why :)

I will add a comment to make it clear

PR #495, opened 2018-12-27
Improve tracing 

geobeau:
Improve tracing by:
- Better choice in the function to trace (less noise from missleading functions)
- More data inserted along opencensus default attributes
- Provides a middleware for tracing (add more data and configuration options)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.14.7] - 2018-12-27
+
+### Improved
+
+- Tracing shows more relevant data in spans (metric names/globs if available)
+- Middleware is provided to improve tracing:
+  - more options to configure when to trace
+  - more data in spans (the executed query and the format)
+
 ## [0.14.6] - 2018-12-18
 
 ### Improved
@@ -451,7 +460,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - We are going to do releases from now on
 
-[Unreleased]: https://github.com/criteo/biggraphite/compare/v0.14.6...HEAD
+[Unreleased]: https://github.com/criteo/biggraphite/compare/v0.14.7...HEAD
+[0.14.7]: https://github.com/criteo/biggraphite/compare/v0.14.6...v0.14.7
 [0.14.6]: https://github.com/criteo/biggraphite/compare/v0.14.5...v0.14.6
 [0.14.5]: https://github.com/criteo/biggraphite/compare/v0.14.4...v0.14.5
 [0.14.4]: https://github.com/criteo/biggraphite/compare/v0.14.3...v0.14.4
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -102,7 +102,8 @@ The list of method type to trace request (default is `["POST", "GET"]`).
 ```python
 BG_TRACING_TARGET_WHITELIST = ["sum(carbon.relays.*local*01*.*)"]
 ```
-A whitelist system to trace only specific requests (set to None and it will trace everything)
+A whitelist system to trace only specific queries (target parameter in graphite's API).
+If it is set to None, it will trace everything.
 
 To configure it, add the middleware to Django's list:
 ```
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -103,3 +103,8 @@ The list of method type to trace request (default is `["POST", "GET"]`).
 BG_TRACING_TARGET_WHITELIST = ["sum(carbon.relays.*local*01*.*)"]
 ```
 A whitelist system to trace only specific requests (set to None and it will trace everything)
+
+To configure it, add the middleware to Django's list:
+```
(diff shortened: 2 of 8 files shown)


biggraphite/plugins/graphite.py
```diff
@@ -25,7 +25,7 @@
 from graphite.logger import log
 from graphite.render import hashing
 
-from biggraphite import accessor as bg_accessor
+from biggraphite import accessor as bg_accessor, tracing
```
adericbourg: I guess this is not used

biggraphite/tracing.py
```diff
@@ -16,19 +16,48 @@
...
+    tracer.tracer = noop_tracer.NoopTracer()
+    tracer.span_context = SpanContext(trace_options=trace_options.TraceOptions(0))
+    tracer.sampler = always_off.AlwaysOffSampler()
+
+def add_attr_to_trace(key, value):
+    """Add an attribute to the current span if tracing is enabled."""
+    if not execution_context:
+        pass
+    _tracer = execution_context.get_opencensus_tracer()
+    _tracer.add_attribute_to_current_span(
```
adericbourg: As the `_tracer` variable is already private to the function, there is no need to prefix it with an underscore.

Same comment for the `trace` function, by the way.

setup.py
```diff
@@ -49,7 +49,7 @@ def _read_reqs(relpath):
 
 setuptools.setup(
     name="biggraphite",
-    version="0.14.6",
+    version="0.14.7",
```
adericbourg: Nitpicking: I would prefer to do version bumps in separate pull-requests.

Otherwise, you may append fix commits to the PR (after the version bump) and it would be misleading to tag the release.

biggraphite/tracing.py
```diff
@@ -16,19 +16,48 @@
...
+
+def stop_trace():
+    """Stop the current trace."""
+    if not execution_context:
+        pass
+    execution_context.set_current_span(None)
+    tracer = execution_context.get_opencensus_tracer()
+    tracer.tracer = noop_tracer.NoopTracer()
+    tracer.span_context = SpanContext(trace_options=trace_options.TraceOptions(0))
+    tracer.sampler = always_off.AlwaysOffSampler()
```
adericbourg: How thread-safe is this block?

What happens if a thread keeps on using the tracer between lines 52 and 54?

CONFIGURATION.md
```diff
@@ -87,4 +87,24 @@ You need to configure integrations to have traces on /render in local_settings.p
...
+
+```python
+BG_TRACING_METHODS = ["POST", "GET"]
+```
+The list of method type to trace request (default is `["POST", "GET"]`).
+
+```python
+BG_TRACING_TARGET_WHITELIST = ["sum(carbon.relays.*local*01*.*)"]
+```
+A whitelist system to trace only specific requests (set to None and it will trace everything)
```
adericbourg: I don't really get what this is. Can you define what a target is?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.14.7] - 2018-12-27
+
+### Improved
+
+- Tracing shows more relevant data
```
adericbourg: What is behind "more relevant data"? Can you be more specific?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Middleware is provided to improve tracing
```
adericbourg: How does it improve tracing?

* If it is worth explaining it in the changelog, please do it.
* If it is not, please remove this from the changelog.

biggraphite/drivers/elasticsearch.py
```diff
@@ -576,10 +574,15 @@ def _search_metrics_from_components(self, glob, components, search=None):
...
 
     @GLOB_METRICS_LATENCY.time()
+    @tracing.trace
     def glob_metrics(self, glob, start_time=None, end_time=None):
         """Return a sorted list of metrics matching this glob."""
         super(_ElasticSearchAccessor, self).glob_metrics(glob, start_time, end_time)
 
+        tracing.add_attr_to_trace("glob", str(glob))
+        tracing.add_attr_to_trace("time.start", str(start_time))
+        tracing.add_attr_to_trace("time.stop", str(end_time))
```
adericbourg: Naming for `time.start` and `time.stop` may be misleading. 

As parameters, they represent the bounds of the search time range. 
As I read them in the tracer, I spontaneously understand that is shows some kind of execution time.

Would you reword this?

_Same for `glob_directory_names`_

setup.py
geobeau: Ack, will rebase before merging to put it on top

biggraphite/tracing.py
geobeau: No thread should share this at this point. To propagate the tracer to another thread it goes through serialization so it will be a copy. If it stills emit a span (which shouldn't happen at this point) it will show as "no root span" on the zipkin/jaeger ui.

PR #498, opened 2019-01-03
fix accessor cache on python 3.x

Roguelazer:
This is the only issue I've found so far running on Python 3.6 instead of Python 2.7 (an upgrade I'm trying to do now since Python 2.7 EOLs this year). Applying this patch results in correct rendering in graphite and correct ingestion through bg-carbon-cache.


adriengentil:
Thanks for your contribution ! Can you rebase before we merge ?

PR #515, opened 2019-02-26
Elasticsearch SCHEMA configurable

melchiormoulin:


jfwm2:
IMHO, it would be way better to be able to set these numbers via the configuration parameters; we are not going to do a PR + merge every time we need to change these settings that do not depend from the implementation and only reflect architectural choice about the platform running biggraphite.

PR #519, opened 2019-03-28
Do not update read_on if the metric is not present

adriengentil:
graphite-web may still have a metric in its cache that have been deleted
in Cassandra. It will re-create the metric partially (every field to null
except name and read_on) when field 'read_on' is getting updated because
UPDATE and INSERT clauses in Cassandra behave like an UPSERT operation.
Adding 'IF EXISTS' makes sure that Cassandra won't create the metric if it
doesn't exists.

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -1254,7 +1254,7 @@ class _CassandraAccessor(bg_accessor.Accessor):
             UPDATE_READ_ON_METADATA,
             __prepare(
                 'UPDATE "%s".metrics_metadata SET read_on=now()'
-                " WHERE name=?;" % self.keyspace_metadata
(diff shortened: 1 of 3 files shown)


iksaif:
Be careful, IF EXISTS might be a bit more costly. (should be fine if only on the read path though)

jfwm2:
Maybe we should add a test (the same kind of the one that should have been there to check that the metric is not partially written with every field to null except name and read_on)

adriengentil:
> Be careful, IF EXISTS might be a bit more costly. (should be fine if only on the read path though)

Tanks for pointing that out! Yes, that would be only on the read path and the update frequency for read_on is configurable.

> Maybe we should add a test (the same kind of the one that should have been there to check that the metric is not partially written with every field to null except name and read_on)

Done

PR #525, opened 2019-05-24
Try to reduce CPU usage on cassandra metadata due to expensive glob

erebe:
Purpose is to cache also queries that ReadTimeout in order to avoid hammering down cassandra metadata cluster

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -1437,6 +1437,7 @@ class _CassandraAccessor(bg_accessor.Accessor):
                 if kwargs.get("raise_on_first_error") is True:
                     raise e
                 result = e
+                result.request = execution_request
                 success = False
             query_results.append((success, result))
         return query_results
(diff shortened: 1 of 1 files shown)


biggraphite/drivers/cassandra.py
```diff
@@ -1899,8 +1899,17 @@ def _extract_results(query_results):
                         if n_metrics > self.max_metrics_per_pattern:
                             raise too_many_metrics
                         yield name
-            except cassandra.DriverException as e:
-                raise CassandraError("Failed to glob: %s on %s" % (table, glob), e)
+                else:
+                    # We want to cache also requests that timeouts in order to avoid re-requesting
```
iksaif: I think it would be better to cache the exception instead of prentending the queries worked on returned nothing

erebe: Ack, 
But i need to test it more, as I am not even convinced myself that we cache things correctly.

For example just above here https://github.com/criteo/biggraphite/pull/525/files#diff-01170f18543de483283e107af3e7991fL1863 If I am understanding correctly
cached_results should contains the metrics names that the query fetched ( as set by https://github.com/criteo/biggraphite/pull/525/files#diff-01170f18543de483283e107af3e7991fL1906) so I don't see how we can remove executions_requests based on this value

PR #543, opened 2020-03-06
Added a way to read directories on ES from a different index

rbizos:
The directory index was added to speed up the directory queries. 
The write is implemented in carbon-relay-ng

### DIFF
--- a/biggraphite/drivers/elasticsearch.py
+++ b/biggraphite/drivers/elasticsearch.py
@@ -160,7 +160,6 @@ def add_argparse_arguments(parser):
 
     parser.add_argument(
         "--elasticsearch_use_directory_index",
-        metavar="NAME",
         help="use secondary index for directories",
         action="store_true",
         default=False,
@@ -690,7 +689,6 @@ class _ElasticSearchAccessor(bg_accessor.Accessor):
         results = [h.name for h in self.glob_metrics(glob, start_time, end_time)]
         return iter(results)
 
-
     @GLOB_DIRECTORY_NAMES_LATENCY.time()
     @tracing.trace
     def glob_directory_names(self, glob, start_time=None, end_time=None):
@@ -1048,7 +1046,6 @@ class _ElasticSearchAccessor(bg_accessor.Accessor):
         if end_time is not None:
             search = search.filter("range", created_on={"lte": end_time.isoformat()})
         return search
-
 
 def build(*args, **kwargs):
     """Return a bg_accessor.Accessor using ElasticSearch."""
--- a/biggraphite/drivers/elasticsearch.py
+++ b/biggraphite/drivers/elasticsearch.py
@@ -35,6 +35,7 @@ from biggraphite import glob_utils as bg_glob
 from biggraphite import metric as bg_metric
 from biggraphite.drivers import _utils
 from biggraphite.drivers import ttls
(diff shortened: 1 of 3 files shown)


biggraphite/drivers/elasticsearch.py
```diff
@@ -958,14 +997,18 @@ def __update_document(self, data, index, document_id):
             ignore=[404, 409],
         )
 
+    def _create_directory_search_query(
```
hdost: Docs what does "index" mean in the context of this function? does it mean that the directory is or is not indexed? consider `is_indexed` if it is a boolean also.

biggraphite/drivers/elasticsearch.py
```diff
@@ -577,6 +584,49 @@ def _search_metrics_from_components(self, glob, components, search=None):
...
 
+    def _search_directory_from_components(self, glob, components, search=None):
+        """Assembles a query to search directory names.
+
+        Raises:
+          InvalidArgumentError: If the components include a globstar
+        """
+        # TODO (r.bizos) add unittest with directory index
+        glob_depth = _get_depth_from_components(components)
+        search = self._create_search_query() if search is None else search
```
hdost: Could be simplified to 
`if search is None: search = self._create_search_query()`

biggraphite/drivers/elasticsearch.py
```diff
@@ -577,6 +584,49 @@ def _search_metrics_from_components(self, glob, components, search=None):
...
+
+        elif self._use_directory_index and self.__glob_parser.is_fully_defined(
+            components[:-1]
+        ):
+            # fully defined parent, only usable with directory index
+            search = search.filter(
+                "term", parent=DIRECTORY_SEPARATOR.join([c[0] for c in components[:-1]])
+            )
+
+        else:
```
hdost: Unnecessary nesting

biggraphite/drivers/elasticsearch.py
rbizos: index is the index used for the search, if not present then it will be set by the method

biggraphite/drivers/elasticsearch.py
```diff
@@ -330,21 +333,25 @@ def __init__(
...
         index_suffix=DEFAULT_INDEX_SUFFIX,
+        directory_index=DEFAULT_DIRECTORY_INDEX,
         username=None,
         password=None,
         timeout=DEFAULT_TIMEOUT,
         updated_on_ttl_sec=ttls.DEFAULT_UPDATED_ON_TTL_SEC,
         read_on_ttl_sec=ttls.DEFAULT_READ_ON_TTL_SEC,
         read_on_sampling_rate=DEFAULT_READ_ON_SAMPLING_RATE,
         schema_path=DEFAULT_ES_SCHEMA_PATH,
+        use_directory_index=False,
```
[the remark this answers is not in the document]
rbizos: Indeed, done

biggraphite/drivers/elasticsearch.py
rbizos: Black formating

biggraphite/drivers/elasticsearch.py
hdost: FAR: It still might be worth documenting the parameters.

PR #560, opened 2021-03-26
Multiple cleaning enhancements & fixes.

mycroft:
Improving cleaning by:

- Adding a way to bypass some offsets automatically if Cassandra not responsive;
- Adding more metrics, especially on cleaning;
- Catch Exceptions on directory cleaning.

### DIFF
--- a/biggraphite/cli/command_clean.py
+++ b/biggraphite/cli/command_clean.py
@@ -72,6 +72,14 @@ class CommandClean(command.BaseCommand):
             default="",
             action="store"
         )
+        parser.add_argument(
+            "--num-token-ignore-on-error",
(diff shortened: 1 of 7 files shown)


biggraphite/cli/command_clean.py
```diff
@@ -72,6 +72,14 @@ def add_arguments(self, parser):
             default="",
             action="store"
         )
+        parser.add_argument(
+            "--num-token-ignore-on-error",
+            help="Specify the number to token to pass to continue cleaning"
```
rbizos, line 77: The default is 0.001 not 0.005 right ? The value is passed as a number not a percentage ?
Also there is a typo `Specify the number of tokens`

PR #561, opened 2021-04-11
Clean expired metrics: New algorithm, tweaks

mycroft:
Adding 2 new ways to clean outdated metrics, and adding several possible tweaks using env. variables.

### DIFF
--- a/biggraphite/drivers/cassandra.py
+++ b/biggraphite/drivers/cassandra.py
@@ -26,6 +26,7 @@ import random
 import time
 from distutils import version
 from os import path as os_path
+from os import environ
 
 import cassandra
 import prometheus_client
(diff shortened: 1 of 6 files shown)


biggraphite/drivers/cassandra.py
```diff
@@ -2596,15 +2597,43 @@ def _clean_expired_metrics(
         cutoff = (int(time.time()) - max_age) * 1000
         log.info("Cleaning with cutoff time %d", cutoff)
 
+        # select method: traditional, unfiltered, adaptative
+        available_methods = ['traditional', 'unfiltered', 'adaptative']
```
camathieu, line 2610: OPT : Add a description of the different algs in the Readme

camathieu:
LGTM

biggraphite/drivers/cassandra.py
```diff
@@ -2505,6 +2505,17 @@ def clean(
 
         first_exception = None
 
+        # 0.006% of database
+        num_token_ignore_on_error = 2 ** 50
+
+        env_num_token = environ.get('CLEAN_IGNORE_BATCH_SIZE_ON_ERROR')
```
hdost, line 2511: I know that the parameter was also changed, but does it make more sense to have this as a parameter than an environment variable?

biggraphite/drivers/cassandra.py
```diff
@@ -2596,15 +2606,51 @@ def _clean_expired_metrics(
         cutoff = (int(time.time()) - max_age) * 1000
         log.info("Cleaning with cutoff time %d", cutoff)
 
+        # select method: traditional, unfiltered, adaptative
+        available_methods = ['traditional', 'unfiltered', 'adaptative']
+        method = 'traditional'
+
+        # use env var to select method.
+        env_method = environ.get('CLEAN_OUTDATED_METHOD')
```
hdost, line 2614: Same comment for this and CLEAN_BATCH_SIZE

biggraphite/drivers/cassandra.py
mycroft: I decided to get rid of the parameter in favor of env. variable because those ones are cassandra-only related; Adding a parameter requires to also change function definitions in hybrid, memory, elasticsearch, and there is no real point of them.  Also, as I really want to fine tweak cassandra cleaning, I might add new configurable settings the same way (and this would make functions & parameter passing really huge).
I'll only need to document those (why default value).

PR #577, opened 2021-06-15
Adding a way to clean up dead metrics records in Cassandra.

mycroft:
In cassandra, there might be some rogue metrics records which does not
have anymore any metrics_metadata records. This new feature, usable
when doing a bgutil repair, will check for each metrics records if
the metrics_metadata counterpart still exists, and if not, delete it.


camathieu:
LGTM

